PR #4, opened 2015-01-12
sepolicy: use ugettext() and unicode() for localized strings

bachradsusi:
When an user uses non-standard localization with utf-8 string, he can get traceback message due use of  8-bit strings instead of unicode strings:

  File "/bin/sepolicy", line 647, in <module>
    sys.stderr.write("%s: %s\n" % (e.__class__.__name__, str(e)))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 48: ordinal not in range(128)

This patch fixes this for sepolicy tool.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1177380


stephensmalley:
Patches for selinux should go to selinux@tycho.nsa.gov.

PR #7, opened 2015-06-04
sepolicy-generate: Handle more reserved port types

masatake:
Currently only reserved_port_t, port_t and hi_reserved_port_t
are handled as special when making a ports-dictionary.
However, as fas as corenetwork.te.in of serefpolicy,
both unreserved_port_t and ephemeral_port_t also handled
in the same way.

Signed-off-by: Masatake YAMATO <yamato@redhat.com>

### DIFF
--- a/policycoreutils/sepolicy/sepolicy/generate.py
+++ b/policycoreutils/sepolicy/sepolicy/generate.py
@@ -90,7 +90,9 @@ def get_rpm_nvr_list(package):
 def get_all_ports():
     dict = {}
     for p in sepolicy.info(sepolicy.PORT):
-        if p['type'] == "reserved_port_t" or \
+        if p['type'] == "ephemeral_port_t" or \
+                p['type'] == "unreserved_port_t" or \
+                p['type'] == "reserved_port_t" or \
                 p['type'] == "port_t" or \
                 p['type'] == "hi_reserved_port_t":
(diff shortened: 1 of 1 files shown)


stephensmalley:
Please post patches to selinux@tycho.nsa.gov (subscribe via email to selinux-join@tycho.nsa.gov).

masatake:
I did.
([PATCH] sepolicy-generate: Handle more reserved port types.)

doverride:
On Wed, Jun 03, 2015 at 08:46:42PM -0700, Masatake YAMATO wrote:
> Currently only reserved_port_t, port_t and hi_reserved_port_t
> are handled as special when making a ports-dictionary.
> However, as fas as corenetwork.te.in of serefpolicy,
> both unreserved_port_t and ephemeral_port_t also handled
> in the same way.
> 

ephemeral port identifiers are a redhat specific thing and i do not believe reference policy has this.

then again, this looks like hard coding customizable identifiers one way or another to me. it sets a less than optimal precedent in that it assume selinux policy is fixed. one of the things that defines selinux in my view is that policy is customisable

masatake:
> ephemeral port identifiers are a redhat specific thing and i do not believe reference policy has this.

Yes. that was a redhat specific.Sorry for making noise.

PR #8, opened 2015-06-11
Fix -Wreturn-type issues.

DanAlbert:
`assert()` only prevents `-Wreturn-type` from firing if asserts are
enabled. Add `__builtin_unreachable()` to make it work with `-UNDEBUG`.


DanAlbert:
On second thought, aborting is probably better.

stephensmalley:
Please submit patches to selinux@tycho.nsa.gov.  Subscribe via selinux-join@tycho.nsa.gov.

PR #9, opened 2015-10-27
Load libsepol.so.1 instead of libsepol.so

bigon:
libsepol.so symlink is usually part of the development package, try to load the library directly instead


stephensmalley:
prefer patches on list

PR #10, opened 2015-10-28
[sandbox] init script of sandbox returns an improper return code in s…

knoha-rh:
…tatus function

Previous sandbox's init script returns 0 only in status function.
This return code causes wrong service status report in system-config-services.

This fix will change the return code to return a proper code for system-config-services.

Signed-off-by: Keigo Noha <knoha@redhat.com>

### DIFF
--- a/policycoreutils/sandbox/sandbox.init
+++ b/policycoreutils/sandbox/sandbox.init
@@ -42,12 +42,15 @@ stop() {
 }
 
 status() {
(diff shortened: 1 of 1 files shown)


stephensmalley:
Prefer patches to be posted to selinux mailing list.

knoha-rh:
Hi.

Thank you for your comment.
I decline this request.

PR #11, opened 2016-02-03
read_spec_entry: fail on non-ascii

shootingatshadow:
Inserting non-ascii characters into a file_contexts file can cause a
failure on labeling but still result in a successful build.

Hard error on non-ascii characters.

Signed-off-by: William Roberts <william.c.roberts@intel.com>


stephensmalley:
Possibly we should simultaneously update process_line() in label_file.h to log an error message if read_spec_entries() returns an error?  Otherwise they likely get no useful output?

shootingatshadow:
Totally agree, the error output on this sucks! Unique return codes perhaps?

shootingatshadow:
stephensmalley, apparently github updates PR's by just updating the forked branch. So I think you have what is up on Android now.

shootingatshadow:
moved to the mailing list

PR #12, opened 2016-03-25
Fix location of selinuxfs mount point

rhatdan:


stephensmalley:
Please post to selinux list too.

PR #22, opened 2016-08-17
fixing incorrect help message

wakeful:
a small fix to correct semanage cli help message.


stephensmalley:
Also seems to be wrong in the semanage.8 man page.
For future reference, we prefer patches to be sent to the selinux mailing list (selinux AT tycho.nsa.gov; subscribe via selinux-join AT tycho.nsa.gov).

PR #35, opened 2016-11-24
fix semanage fcontext help message

cgzones:
the ftype options were incorrect


stephensmalley:
Thanks, will merge since it is trivial, but note that we prefer patches posted to selinux AT tycho.nsa.gov.  Subscribe via selinux-join AT tycho.nsa.gov.

PR #36, opened 2016-11-30
seobject: Handle python error returns correctly

kyle-walker:
Following commit 9792099f (Properly build the swig exception file even if the headers are missing),
the underlying semanage_fcontext_query_local and semanage_fcontext_query calls result in an OSError
return. This results in the following error when attempting to modify a fcontext defined in the
systems base policy.

    libsemanage.dbase_llist_query: could not query record value (No such file or directory).
    OSError: No such file or directory

To resolve the error, handle the OSError exception, but retain the previous query operation.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1398427

Signed-off-by: Kyle Walker <kwalker@redhat.com>

### DIFF
--- a/python/semanage/seobject.py
+++ b/python/semanage/seobject.py
@@ -1953,10 +1953,12 @@ class fcontextRecords(semanageRecords):
             if not exists:
                 raise ValueError(_("File context for %s is not defined") % target)
 
-        (rc, fcontext) = semanage_fcontext_query_local(self.sh, k)
-        if rc < 0:
-            (rc, fcontext) = semanage_fcontext_query(self.sh, k)
-            if rc < 0:
+        try:
+            (rc, fcontext) = semanage_fcontext_query_local(self.sh, k)
+        except OSError:
+            try:
+                (rc, fcontext) = semanage_fcontext_query(self.sh, k)
+            except OSError:
                 raise ValueError(_("Could not query file context for %s") % target)
 
         if setype != "<<none>>":


stephensmalley:
Please post to selinux@tycho.nsa.gov, after subscribing via selinux-join@tycho.nsa.gov.
Prefer patches via mailing list, especially if non-trivial.

stephensmalley:
Also, I don't understand how the referenced commit could be causing this problem (are you sure you are citing the correct one?) and it seems like we ought to fix this in the python bindings rather than the caller(s).

stephensmalley:
Oh, I think I understand now.  commit 9406ace82b12780da84b2553cb74f88101978ea2 changed the python bindings to throw exceptions rather than return NULL, so that is the actual cause.  However, until commit 9792099fd7847266377df151e7738f9b38ffc18d, you likely weren't getting the exception wrappers at all because you didn't have the headers installed in the system path, and thus we didn't see this behavior until that commit.

kyle-walker:
@stephensmalley 
Correct! Sorry for the incorrect submit, and the missing detail above. You are absolutely correct. The issue only recently came to light because 9792099 was included in a rebase and fixed the silent compile-time failure.

Would it be beneficial to close this PR and resubmit to the mailing list instead?

stephensmalley:
Yes, we'd prefer that.

kyle-walker:
Absolutely, closing now and will pursue the mailing list posting.

Thank you for the patience and guidance!

PR #52, opened 2017-04-03
libselinux: PYSITEDIR has dup path

matthew-l-weber:
When $(DESTDIR) is defined, the $(PYSITEDIR) ends up with a double
absolute path.  (In the normal case with $(DESTDIR) not defined,
site.getsitepackages()[0] already results in an absolute path)

Signed-off-by: Matthew Weber <matthew.weber@rockwellcollins.com>


matthew-l-weber:
See issue ticket above for non-issue status after analysis.

PR #59, opened 2017-06-19
Portability patches

pks-t:
This is a first batch of small portability fixes. It removes a few instances of non-portable functions, problems with hardened toolchains defining FORTIFY_SOURCE and removes usage of __BEGIN_DECLS.

While I guess that the first two commits of removing __BEGIN_DECLS and the patch for FORTIFY_SOURCE are uncontroversial, I can imagine the removal of getpwent_r being a bit more so due to the alternative being non-reentrant. But to the best of my knowledge the code in question is not even called in a threaded context, so I actually doubt it matters much. Anyway, seeing that I'm new to this code base I could very likely be wrong.


stephensmalley:
Prefer patches submitted via git send-email or equivalent to selinux AT tycho.nsa.gov unless unable to do so; subscribe via selinux-join AT tycho.nsa.gov.

pks-t:
Will do, thanks for the info!

PR #60, opened 2017-07-18
Fixed bad reference in roleattribute

gcolbert:
"roleattribute" was referencing "typeattributeset", but I believe it should be referencing "roleattributeset" instead.


stephensmalley:
Prefer patches via git send-email to selinux AT tycho.nsa.gov, subscribe via selinux-join AT tycho.nsa.gov.

jwcart2:
I have applied this change.

PR #67, opened 2017-10-19
selinux: libselinux: Enable multiple input files to selabel_open.

stephensmalley:
The file_contexts labeling backend, specified in label_file.c, currently assumes
that only one path will be specified as an option to selabel_open().  The split
of platform and non-platform policy on device, however, will necessitate the
loading of two disparate policy files.  Rather than combining the files and then
calling the existing API on a newly-formed file, just add the ability to specify
multiple files to use.  Order of opt specification to selabel_open matters.

This corresponds to AOSP commit 50400d38203e4db08314168e60c281cc61a717a8, which
lead to a fork with upstream, which we'd like to correct.

Signed-off-by: Dan Cashman <dcashman@android.com>

### DIFF
--- a/libselinux/src/label.c
+++ b/libselinux/src/label.c
@@ -143,7 +143,11 @@ static int selabel_fini(struct selabel_handle *rec,
 			    struct selabel_lookup_rec *lr,
 			    int translating)
 {
-	if (compat_validate(rec, lr, rec->spec_file, 0))
+	char *path = NULL;
+
+	if (rec->spec_files)
+		path = rec->spec_files[0];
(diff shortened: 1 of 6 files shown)


stephensmalley:
Do not merge!  Just testing whether our travis-ci tests would catch the fact that this commit introduces a bug that produces seg faults in coreutils, selabel_lookup, and other userspace programs.

stephensmalley:
Ok, apparently not. Guess we need to look at what needs to be added to the travis-ci tests.

williamcroberts:
Very cool, that answers my question I posted to you.

PR #75, opened 2018-01-16
Rework of Makefiles

stephensmalley:
I have reworked the Makefiles a bit to make it more compatible with
the standard semantics for DESTDIR and PREFIX.

What made me look at this was a package I was working on in 
Buildroot[1] that was using the pkg-config file generated
from libselinux. The pkg-config file generated has bogus paths 
which results in that my package did not compile.

When looking further in our packages for libselinux/libsepol/libsemanage,
we allready do "tricks" to make it compile.
For example, we set DESTDIR during compilation to compute library and header paths.

DESTDIR should be supported only in the install* and uninstall* targets,
as those are the only targets where it is useful. [2]

PREFIX should be used to set a prefix for the installation.
This could be provided both in compile and install stages.

For example, consider the following line:
#make DESTDIR=/tmp/selinux install

It will generate a libselinux.pc file that looks like this:

prefix=/tmp/selinux/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=/tmp/selinux/usr/include

The DESTDIR should not be part of the prefix. The correct output should be:

prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=/usr/include

The following example do not compile at all:
#make DESTDIR=/tmp/myroot PREFIX=/selinux install

In this case, everything should be installed under /tmp/myroot/selinux.
The prefix in all generated pkg-files should have the path /selinux/xxxxx ( not /tmp/myroot/selinux/xxxx).

This patchsets tries to solve the following issues:
- The pkg-config files generates odd paths when using DESTDIR without PREFIX
- DESTDIR is needed during compile time to compute library and header paths which it should not.
- Installing with both DESTDIR and PREFIX set gives us odd paths
- Make usage of DESTDIR and PREFIX more standard

However, DESTDIR is still needed in the "root" Makefile to make libsemanage to compile.
But all components (libselinux, libsepol and libsemanage) can be compiled separatly with
standard usage of DESTDIR and PREFIX.

Thanks,

Best regards
Marcus Folkesson

[1] https://buildroot.org/
[2] https://www.gnu.org/prep/standards/html_node/DESTDIR.html

### DIFF
--- a/README
+++ b/README
@@ -19,6 +19,6 @@ lacks library functions or other dependencies relied upon by your
 distribution.  If it breaks, you get to keep both pieces.
 
 To install libsepol on macOS (mainly for policy analysis):
-cd libsepol; make DESTDIR=/usr/local PREFIX=/usr/local install
+cd libsepol; make PREFIX=/usr/local install
 
 This requires GNU coreutils (brew install coreutils).
--- a/libselinux/include/Makefile
+++ b/libselinux/include/Makefile
@@ -1,6 +1,6 @@
 # Installation directories.
-PREFIX ?= $(DESTDIR)/usr
-INCDIR ?= $(PREFIX)/include/selinux
+PREFIX ?= /usr
+INCDIR = $(DESTDIR)$(PREFIX)/include/selinux
 
 all:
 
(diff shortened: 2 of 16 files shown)


stephensmalley:
Alters install location of the shared objects.

PR #76, opened 2018-01-25
libsemanage: Allow tmp files to be kept if a compile fails

williamcroberts:
Allow the tmp build files to be kept for debugging when a policy
build fails.

Signed-off-by: Richard Haines <richard_c_haines@btinternet.com>


williamcroberts:
Just putting this up to run the CI on it, not ready for merge.

PR #77, opened 2018-01-25
gui: remove selinux-sepolgengui

fishilico:
Hi,
I sent a few hours ago these two patches on the mailing list, by as the first one seems to be blocked somewhere (I have only received back the second one), I am publishing them on Github too, as a Pull Request.

These patches removes selinux-sepolgengui because this application is not compatible with Gtk3, Python 3 (it requires PyGTK), etc. and would otherwise require some effort to update it.
More precisely, even though ``pygi-convert.sh`` updated the code in order to try making the application compatible with Python 3, PyGI... (in commit 0f3beeb00e7a42cc2f44ef0392b8a3a7566a17d7), ``polgen.glade`` is still incompatible with Gtk-Builder (it would need to be converted). I do not want to spend time converting this file when I see that a bug preventing this application to launch has been present for more than one year.

Cheers


fishilico:
Petr Lautrbach have submitted patches which fix sepolgengui: https://marc.info/?l=selinux&m=151860211514842&w=2 . As this component is used, I drop this pull request.

PR #82, opened 2018-02-28
libsemanage: Improve warning for installing disabled module

williamcroberts:
Resolves: rhbz#1337199

Signed-off-by: Vit Mojzis <vmojzis@redhat.com>


williamcroberts:
One of those fun random CURL failures, merging.

PR #83, opened 2018-03-09
WIP: Delete identical genfscon-s

phhusson:
secilc has a multiple_decls option to allow for multiple type
declarations.
The next step is to allow multiple samples of the same rules.
This commit does this on genfscon

One usecase is Android/Project Treble:
With Project Treble, vendor might include rules included in later
in framework.
In order to be able to update the framework in this case, we need
to remove identical rules.

I have several pending questions before considering merging:
- Should the "compact" function be somewhere else? Or perhaps there is already some variant available?
- Should the "compact" function simply take a cil_sort rather than a C array?
- Should we compact all types indifferently? If so, we need to guarantee that the _compare function returns 0 only when the types rules are identical, and not just the same match rule. Is this already the case?
- How is memory allocation done/will compact impact the release of the memory?


stephensmalley:
@jwcart2 please take to selinux mailing list for discussion.  Subscribe via selinux-join@tycho.nsa.gov.

PR #85, opened 2018-03-09
libsepol: Export sepol_polcap_getnum/name functions

stephensmalley:
Export the sepol_polcap_getnum/name() functions to users of
the shared library.  This will enable SETools to stop depending
on the static library.

Note that we may want to move polcaps.h up one level since
the convention is that headers directly under include/sepol are
shared library APIs while headers under include/sepol/policydb
are limited to static users.  However, this will unnecessarily
break the build for existing static users so it is deferred.

Suggested-by: Chris PeBenito <pebenito@ieee.org>


williamcroberts:
I just prepping this patch, I see you have it.

PR #86, opened 2018-03-14
libselinux, libsemanage: Replace PYSITEDIR with PYTHONLIBDIR

bachradsusi:
libselinux and libsemanage Makefiles invoke site.getsitepackages() in
order to get the path to the directory /usr/lib/pythonX.Y/site-packages
that matches the Python interpreter chosen with $(PYTHON). This method
is incompatible with Python virtual environments, as described in
https://github.com/pypa/virtualenv/issues/355#issuecomment-10250452 .
This issue has been opened for more than 5 years.

On the contrary python/semanage/ and python/sepolgen/ Makefiles use
distutils.sysconfig.get_python_lib() in order to get the site-packages
path into a variable named PYTHONLIBDIR. This way of computing
PYTHONLIBDIR is compatible with virtual environments and gives the same
result as PYSITEDIR.

As PYTHONLIBDIR works in more cases than PYSITEDIR, make libselinux and
libsemanage Makefiles use it. And as native code is installed (as part
of the SWIG wrapper), use "plat_specific=1" in order to use /usr/lib64
on systems which distinguish /usr/lib64 from /usr/lib.

Signed-off-by: Nicolas Iooss <nicolas.iooss@m4x.org>


fishilico:
Merged in 07629c0a9f26e6e9ebb0b2847e09a461512990c9

PR #101, opened 2018-10-03
restorecond: Do not ignore the -f option

WOnder93:
Since the default value of watch_file is set unconditionally *after* the
command-line arguments have been parsed, the -f option is (and has
always been) effectively ignored. Fix this by setting it before the
parsing.

Signed-off-by: Ondrej Mosnacek <omosnace@redhat.com>


WOnder93:
Now posted to the list, as it should have been, sorry: https://www.mail-archive.com/selinux@tycho.nsa.gov/msg07835.html

PR #103, opened 2018-10-11
libsepol: Add ability to sort ocontexts in libsepol and add option to use it in checkpolicy

williamcroberts:
ocontexts (initial sids, fs_use_*, genfscon, portcon, etc) are sorted by libsemanage when using policy modules and by libsepol when using CIL, but they are not sorted by checkpolicy when creating a policy from a policy.conf.

Checkpolicy's behavior allows control over the ordering which determines the matching order for portcons and other ocontext rules, but there are times when that specific control is not desired.

This patch set exposes an internal ocontext sorting function and adds a command line option to checkpolicy to sort ocontexts.

James Carter (2):
  libsepol: Create policydb_sort_ocontexts()
  checkpolicy: Add option to sort ocontexts when creating a binary
    policy

 checkpolicy/checkpolicy.c                  | 22 +++++++++++++++++-----
 libsepol/include/sepol/policydb/policydb.h |  2 ++
 libsepol/src/policydb.c                    |  5 +++++
 3 files changed, 24 insertions(+), 5 deletions(-)

### DIFF
--- a/checkpolicy/checkpolicy.c
+++ b/checkpolicy/checkpolicy.c
@@ -111,9 +111,9 @@ unsigned int policyvers = POLICYDB_VERSION_MAX;
 static __attribute__((__noreturn__)) void usage(const char *progname)
 {
 	printf
-	    ("usage:  %s [-b[F]] [-C] [-d] [-U handle_unknown (allow,deny,reject)] [-M]"
-	     "[-c policyvers (%d-%d)] [-o output_file] [-t target_platform (selinux,xen)]"
-	     "[input_file]\n",
+	    ("usage:  %s [-b[F]] [-C] [-d] [-U handle_unknown (allow,deny,reject)] [-M] "
+	     "[-c policyvers (%d-%d)] [-o output_file] [-S] "
+	     "[-t target_platform (selinux,xen)] [input_file]\n",
 	     progname, POLICYDB_VERSION_MIN, POLICYDB_VERSION_MAX);
 	exit(1);
 }
@@ -394,7 +394,7 @@ int main(int argc, char **argv)
 	size_t scontext_len, pathlen;
 	unsigned int i;
 	unsigned int protocol, port;
-	unsigned int binary = 0, debug = 0, cil = 0, conf = 0;
+	unsigned int binary = 0, debug = 0, sort = 0, cil = 0, conf = 0;
 	struct val_to_name v;
 	int ret, ch, fd, target = SEPOL_TARGET_SELINUX;
 	unsigned int nel, uret;
@@ -418,11 +418,12 @@ int main(int argc, char **argv)
 		{"mls", no_argument, NULL, 'M'},
 		{"cil", no_argument, NULL, 'C'},
 		{"conf",no_argument, NULL, 'F'},
+		{"sort", no_argument, NULL, 'S'},
 		{"help", no_argument, NULL, 'h'},
 		{NULL, 0, NULL, 0}
 	};
 
-	while ((ch = getopt_long(argc, argv, "o:t:dbU:MCFVc:h", long_options, NULL)) != -1) {
+	while ((ch = getopt_long(argc, argv, "o:t:dbU:MCFSVc:h", long_options, NULL)) != -1) {
 		switch (ch) {
 		case 'o':
 			outfile = optarg;
@@ -462,6 +463,9 @@ int main(int argc, char **argv)
 				break;
 			}
 			usage(argv[0]);
+		case 'S':
+			sort = 1;
+			break;
 		case 'M':
 			mlspol = 1;
 			break;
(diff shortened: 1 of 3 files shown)


StidOfficial:
I have this error, during compilation :

```
checkpolicy.c: In function 'main':
checkpolicy.c:645:12: error: implicit declaration of function 'policydb_sort_ocontexts'; did you mean 'policydb_user_cache'? [-Werror=implicit-function-declaration]
      ret = policydb_sort_ocontexts(&policydb);
            ^~~~~~~~~~~~~~~~~~~~~~~
            policydb_user_cache
cc1: all warnings being treated as errors
```

fishilico:
This error occurs for example if checkpolicy is been built with a libsepol version which is not master (this PR also introduced `policydb_sort_ocontexts` in `libsepol/include/sepol/policydb/policydb.h`). Is it what you are doing? Which version of libsepol are you using? You may need to add a `-I/path/to/libsepol/include` to the `CFLAGS` you are using.

StidOfficial:
Try to compile the master branch (`make -C selinux all`), and yes not includes is set.

```
cc -O2 -Werror -Wall -Wextra -Wmissing-format-attribute -Wmissing-noreturn -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -o checkpolicy.o -c checkpolicy.c
```

fishilico:
The way the tools and libraries are packaged (each subdirectory in its own package) makes it difficult to do things like including `../libsepol/include` from `checkpolicy/`. In order to build the current `master` branch, the supported way consists in using `make DESTDIR=... install`, with `DESTDIR` targeting a directory where include files and libraries get copied and used by the tools. This is handled by https://github.com/SELinuxProject/selinux/blob/d3c359ee82e6ac0b18f4db98da726302e69862df/Makefile#L22-L31

Once the libraries are installed there, `make DESTDIR=... all` (and `make DESTDIR=... test`) should work.

StidOfficial:
I have this error with `make DESTDIR=/tmp/test install` :

```
run_init.c: In function 'authenticate_via_shadow_passwd':
run_init.c:210:25: error: implicit declaration of function 'crypt'; did you mean 'creat'? [-Werror=implicit-function-declaration]
  encrypted_password_s = crypt(unencrypted_password_s,
                         ^~~~~
                         creat
run_init.c:210:23: error: assignment to 'char *' from 'int' makes pointer from integer without a cast [-Werror=int-conversion]
  encrypted_password_s = crypt(unencrypted_password_s,
                       ^
cc1: all warnings being treated as errors
```

Is not possible to add like this ?
```
diff --git a/checkpolicy/Makefile b/checkpolicy/Makefile
index 4c817cd6..1fadd876 100644
--- a/checkpolicy/Makefile
+++ b/checkpolicy/Makefile
@@ -9,13 +9,13 @@ TARGETS = checkpolicy checkmodule
 LEX = flex
 YACC = bison -y
 
-CFLAGS ?= -g -Wall -Werror -Wshadow -O2 -pipe -fno-strict-aliasing
+CFLAGS ?= -g -Wall -Werror -Wshadow -O2 -pipe -fno-strict-aliasing -I../libsepol/include
 
 # If no specific libsepol.a is specified, fall back on LDFLAGS search path
 # Otherwise, as $(LIBSEPOLA) already appears in the dependencies, there
 # is no need to define a value for LDLIBS_LIBSEPOLA
 ifeq ($(LIBSEPOLA),)
-        LDLIBS_LIBSEPOLA := -l:libsepol.a
+        LDLIBS_LIBSEPOLA := ../libsepol/src/libsepol.a
 endif
 
 CHECKOBJS = y.tab.o lex.yy.o queue.o module_compiler.o parse_util.o \
diff --git a/checkpolicy/test/Makefile b/checkpolicy/test/Makefile
index 89e7557c..6568a237 100644
--- a/checkpolicy/test/Makefile
+++ b/checkpolicy/test/Makefile
@@ -7,7 +7,7 @@ CFLAGS ?= -g -Wall -W -Werror -O2 -pipe
 # Otherwise, as $(LIBSEPOLA) already appears in the dependencies, there
 # is no need to define a value for LDLIBS_LIBSEPOLA
 ifeq ($(LIBSEPOLA),)
-        LDLIBS_LIBSEPOLA := -l:libsepol.a
+        LDLIBS_LIBSEPOLA := ../../libsepol/src/libsepol.a
 endif
 
 all: dispol dismod
```

fishilico:
Using `../libsepol/src/libsepol.a` would force directory `libsepol` to be present every time `checkpolicy` is built. This is the case when working on this git repository, but not when building the packages (when building `checkpolicy`, `libsepol` may be installed without its source files being available). This is why there is this `-l:libsepol.a` notation, which looks for `libsepol.a` in directories specified by `-L` in `$LDFLAGS`, which contains `-L$(DESTDIR)$(PREFIX)/lib` when building from the root directory.

For the issue with `crypt`, please report this issue separately, and please give some information on the system you are using (at least OS/distribution, compiler name and version, libc name and version). `man crypt` tells:

    #define _XOPEN_SOURCE       /* See feature_test_macros(7) */
    #include <unistd.h>

... and `run_init.c` includes `unistd.h` with this macro, defined in `policycoreutils/run_init/Makefile`. Therefore it seems to be likely a problem with your libc not following the standard which is documented in the man page, or with the `Makefile` not applying `CFLAGS += -D_XOPEN_SOURCE=500`.

PR #113, opened 2018-11-07
dbus: remove deprecated at_console statement

teg:
As described in [0], this likely did not have the intended effect, so
simply remove it. The change in behavior is that up until this patch
it would be possible for any non-system user to potentially gain access
to selinux' dbus interface. Now this is extended to also allow any
system user.

As the comment indicates, PolicyKit is used to enforce access, so this
should be perfectly harmless.

[0]: <https://www.spinics.net/lists/linux-bluetooth/msg75267.html>

Signed-off-by: Tom Gundersen <teg@jklm.no>
CC: David Herrmann <dh.herrmann@gmail.com>

### DIFF
--- a/dbus/org.selinux.conf
+++ b/dbus/org.selinux.conf
@@ -12,12 +12,8 @@
 
   <!-- Allow anyone to invoke methods on the interfaces,
        authorization is performed by PolicyKit -->
-  <policy at_console="true">
-    <allow send_destination="org.selinux"/>
-  </policy>
   <policy context="default">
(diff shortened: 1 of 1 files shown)


bachradsusi:
I'm reopening this ticket so that CI results are available. It will be closed when the patch is merged.

bachradsusi:
https://github.com/SELinuxProject/selinux/commit/3a40d0573556c176da81b6daa1c6ce97a47a8b12

Issue #81, opened 2018-02-26
[semanage] [python-sepol] semanage crashes on systems without policy


schmittlauch:
The semanage tool crashes on systems without installed policies:

```
# semanage
Traceback (most recent call last):
  File "/usr/sbin/semanage", line 28, in <module>
    import seobject
  File "/usr/lib/python3.4/site-packages/seobject.py", line 1039, in <module>
    class portRecords(semanageRecords):
  File "/usr/lib/python3.4/site-packages/seobject.py", line 1041, in portRecords
    valid_types = list(list(sepolicy.info(sepolicy.ATTRIBUTE, "port_type"))[0]["types"])
  File "/usr/lib/python3.4/site-packages/sepolicy/__init__.py", line 196, in <genexpr>
    return ({
  File "/usr/lib/python3.4/site-packages/setools/typeattrquery.py", line 65, in results
    for attr in self.policy.typeattributes():
AttributeError: 'NoneType' object has no attribute 'typeattributes'
```

This is caused by sepol (the python module) setting the global _pol variable to `None` – something setools.TypeAttributeQuery can't deal with.

### platform details:

python3-modules used with python-3.4.3 
distribution: mer
libselinux/ libsepol/ libsemanage/ policycoreutils: v2.7
python3-setools: 4.1.1

disclaimer: policycoreutils and setools are customly packaged, this is a possible (but unlikely) error source

stephensmalley:
Reproduced upstream, e.g. just mv /etc/selinux /etc/selinux.old and try running semanage as above.

fishilico:
This behavior is mainly due to having `valid_types = list(list(sepolicy.info(...` in class definitions instead of having them in class constructors (in `seobject.py`). This makes it harder to catch the fact that no policy is loaded.
I suggest moving the `valid_types` definitions to class constructors, to give importers of seobject.py the possibility to display a nice error message if no policy is loaded (or to specify a policy to be loaded).
What do you think of this?

bachradsusi:
I should probably drop the reference to this issue from my working commits in order to prevent another automatic closing. Sorry for the noise.

bachradsusi:
Fixed by https://github.com/SELinuxProject/selinux/commit/a73b0bba1a52adec3cfe654f726388a68b73e2d3 https://github.com/SELinuxProject/selinux/commit/e718c2ab77f8f3347bd2289502797bb978c33e3e

devZer0:
[...]

PR #116, opened 2018-12-20
semanage without default policy; and with -S

bachradsusi:
The following set of patches tries to fix semanage in non-standard environments.

A backtrace when semanage was run on a system without installed policy, see
https://github.com/SELinuxProject/selinux/issues/81
https://bugzilla.redhat.com/show_bug.cgi?id=1558861 

The fix is based on
https://github.com/fishilico/selinux/commit/00eb0cc86185d2145ea3ec61df08c587faba1375
by Nicolas Iooss <nicolas.iooss@m4x.org>

And when semanage was run -S <store>, sepolicy module was initialized with the
default policy file (targeted) and also selinux module used the same default. It
meant that semanage either backtraced when the default policy was not installed
or used wrong store and therefore didn't behave correctly.

E.g. # semanage login -S minimum -d -s unconfined_u -r s0-s0:c0.c1023 __default__
libsemanage.dbase_llist_query: could not query record value
OSError: [Errno 0] Error


fishilico:
These commits have been merged as https://github.com/SELinuxProject/selinux/commits/1015aef5cfa6a88399265385040ed584e00d363e . Thanks!

Issue #121, opened 2019-01-02
errno should be reset to 0 before each call to getpwent() in genhomedircon


bigon:
Hello,

In `genhomedircon`, it seems that `errno` should be reset to 0 before each call  to `getpwent()`

The manpage of `getpwent()` states:

> The  getpwent()  function  returns a pointer to a passwd structure, or NULL if there are no more entries or an error occurred.  If an error occurs, errno is set appropriately.  If one wants to check errno after the call, it should be set to zero before the call.

Without this, `genhomedircon` can wrongly return the following error when using the `files` module:

> libsemanage.get_home_dirs: Error while fetching users.  Returning list so far.

fishilico:
The catch is this issue is that `getpwent` can set `errno = EINVAL` when returning entries, which is why `errno` needs to be set to 0 every time `getpwent` is called.

For future reference, here is a simple test program to illustrate this behavior:
```C
#include <sys/types.h>
#include <pwd.h>
#include <errno.h>
#include <stdio.h>

int main(void)
{
    struct passwd *pw;
    setpwent();
    errno = 0;
    while ((pw = getpwent()) != NULL) {
        printf("%d: %s (errno %d)\n", pw->pw_uid, pw->pw_name, errno);
	errno = 0;
    }
    printf("getpwent errno = %d (%m)\n", errno);
    endpwent();
    return 0;
}
```

Its output (in a Vagrant virtual machine with glibc 2.28 and systemd 239.370 is:
```
0: root (errno 22)
1: bin (errno 22)
2: daemon (errno 22)
8: mail (errno 22)
14: ftp (errno 22)
33: http (errno 22)
68: uuidd (errno 22)
81: dbus (errno 22)
99: nobody (errno 22)
191: systemd-journal-gateway (errno 22)
192: systemd-timesync (errno 22)
193: systemd-network (errno 22)
194: systemd-bus-proxy (errno 22)
195: systemd-resolve (errno 22)
998: systemd-journal-remote (errno 22)
999: systemd-coredump (errno 22)
997: systemd-journal-upload (errno 22)
1000: vagrant (errno 22)
994: git (errno 22)
993: avahi (errno 22)
992: colord (errno 22)
102: polkitd (errno 22)
getpwent errno = 0 (Success)
```

I merged your commit. Thanks!

PR #122, opened 2019-01-02
libsemanage: Always set errno to 0 before calling getpwent()

bigon:
The manpage explicitly states that:

  The  getpwent()  function  returns a pointer to a passwd structure, or
  NULL if there are no more entries or an error occurred.  If an error
  occurs, errno is set appropriately.  If one wants to check errno after
  the call, it should be set to zero before the call.

Without this, genhomedircon can wrongly return the following:
  libsemanage.get_home_dirs: Error while fetching users.  Returning list so far.

https://github.com/SELinuxProject/selinux/issues/121

Signed-off-by: Laurent Bigonville <bigon@bigon.be>

### DIFF
--- a/libsemanage/src/genhomedircon.c
+++ b/libsemanage/src/genhomedircon.c
@@ -361,7 +361,11 @@ static semanage_list_t *get_home_dirs(genhomedircon_settings_t * s)
 
 	errno = 0;
 	setpwent();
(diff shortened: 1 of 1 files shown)


fishilico:
Merged as https://github.com/SELinuxProject/selinux/commit/9ac345e8d5b1f661b1012380d648d73d933e503a. Thanks

Issue #109, opened 2018-11-03
"semodule -p" accesses files on root filesystem


ensc:
I want to try build offline an SELinux policy and prepared a filesystem like

```
  ~/src/policy-root
  |--- etc/selinux/selinux/...
  |--- var/lib/selinux/minimum/...
```

Most operations work fine; e.g.

```
semodule -p ~/src/policy-root -s 'minimum' -n -i my-module.pp
semodule -p ~/src/policy-root -s 'minimum' -n -B
```

builds a (working) policy.  But it appears, the `semodule` still accesses files on the rootfs. E.g. running it through `strace` shows:

```
12457 openat(AT_FDCWD, "/etc/selinux/config", O_RDONLY|O_CLOEXEC) = 3
12457 access("/sbin/load_policy", X_OK) = 0
....
12457 openat(AT_FDCWD, "...../policy-root/var/lib/selinux/minimum/tmp/booleans.local", O_RDONLY) = 4
12457 openat(AT_FDCWD, "/etc/selinux/targeted/booleans.subs_dist", O_RDONLY|O_CLOEXEC) = 5

```

At least the last part seems to cause problems because substitution rules from host are applied to the offline policy.  Here, it removes e.g.  `allow_mount_anyfile=1` from the `booleans.local` file because host configuration contains

```
allow_mount_anyfile mount_anyfile
```

which does not exist for the offline policy.

stephensmalley:
The problem here seems to be that libsemanage is using a libselinux interface, selinux_boolean_sub(), that internally accesses the "active" policy files, without having previously directed libselinux to an alternate root via selinux_set_policy_root().  Simplest fix would likely be to save, set, and restore the libselinux policy root around the selinux_boolean_sub() call.  More complete approach would be to call selinux_set_policy_root() when setting the semanage root but this will require changes elsewhere to retain the ability to detect when we are installing to the active policy store.

stephensmalley:
See if the PR fixes your issue

PR #127, opened 2019-01-07
libsemanage: set selinux policy root around calls to selinux_boolean_sub

stephensmalley:
As reported in #109, semodule -p /path/to/policyroot -s minimum -n -B
tries to use /etc/selinux/targeted/booleans.subs_dist.  This is because
it invokes the libselinux selinux_boolean_sub() interface, which uses
the active/installed policy files rather than the libsemanage ones.

Switch the selinux policy root around the selinux_boolean_sub() call
to incorporate the semanage root as a prefix and to use the specified
policy store as a suffix so that the correct booleans.subs_dist file
(if any) is used.

The underlying bug is that booleans.subs_dist is not itself managed
via libsemanage. If it was managed and therefore lived within the
policy store, then libsemanage could access the appropriate
booleans.subs_dist file without using the libselinux interface at all,
and thus would not need to modify the selinux policy root.  Moving
booleans.subs_dist to a managed file is deferred to a future change.

Test:
strace semodule -p ~/policy-root -s minimum -n -B

Before:
openat(AT_FDCWD, "/etc/selinux/targeted/booleans.subs_dist", O_RDONLY|O_CLOEXEC) = 5

After:
openat(AT_FDCWD, "/home/sds/policy-root/etc/selinux/minimum/booleans.subs_dist", O_RDONLY|O_CLOEXEC) = 5

Fixes https://github.com/SELinuxProject/selinux/issues/109

### DIFF
--- a/libsemanage/src/boolean_record.c
+++ b/libsemanage/src/boolean_record.c
@@ -6,7 +6,9 @@
  * Implements: record_key_t (Database Record Key)
  */
 
+#include <string.h>
 #include <sepol/boolean_record.h>
+#include "handle_internal.h"
 
 typedef sepol_bool_t semanage_bool_t;
 typedef sepol_bool_key_t semanage_bool_key_t;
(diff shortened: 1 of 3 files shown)


libsemanage/src/boolean_record.c
```diff
@@ -84,10 +85,49 @@ hidden_def(semanage_bool_get_name)
...
+		goto out;
+	end = strrchr(olddir, '/');
+	if (!end)
+		goto out;
+	end++;
+	*end = '\0';
+	rc = asprintf(&newroot, "%s%s%s", prefix, olddir, storename);
+	if (rc < 0)
+		goto out;
+	rc = selinux_set_policy_root(newroot);
```
stephensmalley: I could skip the setting of the policy root here and resetting it below if !strcmp(oldroot, newroot).  Not sure that is worth it though...

fishilico:
Merged as dc3d9c6d7173354cc4caf225ce75c41239cf47bd.

Issue #123, opened 2019-01-03
RFE: setsebool: add option to force even if SELinux detected as disabled


dustymabe:
Hi 

Over in https://github.com/projectatomic/rpm-ostree/issues/1634#issuecomment-435961951 we have identified a need to be able to force setsebool to continue even if selinux appears to be disabled.

Quoting Colin from the comment:

*The quick hack I can think of is an environment variable like e.g. SEMANAGE_FORCE_BUILD=offline or so that was interpreted by setsebool and convinced the tool to rebuild policy and set booleans even if it appeared that SELinux was disabled.*

Would it be acceptable to implement such a configuration ENV variable for setsebool?

stephensmalley:
Not sure offhand why setsebool aborts immediately if SELinux is disabled as it could at least apply persistent changes (assuming -P).  On the other hand, I was wondering semanage boolean provides a way to do this already that doesn't depend on SELinux enabled?

bachradsusi:
In this case, semanage would bring unnecessary dependencies into the system and setsebool already uses libsemanage for setting booleans. It looks like the check whether SELinux is enabled is not necessary; or should not be used when -N is used.

dustymabe:
> Not sure offhand why setsebool aborts immediately if SELinux is disabled as it could at least apply persistent changes (assuming -P). On the other hand, I was wondering semanage boolean provides a way to do this already that doesn't depend on SELinux enabled?

I did not know about `semanage boolean`. Thanks! In our particular use case the `setsebool` operations are executing in rpm post scripts so we can't easily switch to `semanage boolean` without changing a bunch of rpms that we don't control. If possible it would be great if we could get some feature/fix into `setsebool` for this.

stephensmalley:
Tested semanage boolean on a SELinux-disabled system and that worked without problems.
Removed the is_selinux_enabled() test from setsebool and it still failed later on on a SELinux-disabled system, so that requires further investigation into how it differs from semanage boolean and why.

stephensmalley:
setsebool doesn't provide a way to specify the policy store, so it presumes that you want to act on the "active" policy, as configured by /etc/selinux/config.  So even if it were to "work" on a SELinux-disabled system, it would presume that whatever /etc/selinux/config has as the SELINUXTYPE is the policy it should modify.  Is that a safe assumption here?

dustymabe:
[...]

PR #129, opened 2019-01-10
setsebool: support use of -P on SELinux-disabled hosts

stephensmalley:
As reported in #123, setsebool immediately exits with an error if
SELinux is disabled, preventing its use for setting boolean persistent
values.  In contrast, semanage boolean -m works on SELinux-disabled
hosts.  Change setsebool so that it can be used with the -P option
(persistent changes) even if SELinux is disabled.  In the SELinux-disabled
case, disable the policy reload and skip setting of active boolean
values, but set the persistent value in the policy store.

Fixes: https://github.com/SELinuxProject/selinux/issues/123

### DIFF
--- a/policycoreutils/setsebool/setsebool.c
+++ b/policycoreutils/setsebool/setsebool.c
@@ -18,7 +18,7 @@
 #include <errno.h>
 
 int permanent = 0;
-int reload = 1;
+int no_reload = 0;
 int verbose = 0;
 
 int setbool(char **list, size_t start, size_t end);
@@ -38,11 +38,6 @@ int main(int argc, char **argv)
 	if (argc < 2)
 		usage();
 
-	if (is_selinux_enabled() <= 0) {
-		fputs("setsebool:  SELinux is disabled.\n", stderr);
-		return 1;
-	}
-
 	while (1) {
 		clflag = getopt(argc, argv, "PNV");
 		if (clflag == -1)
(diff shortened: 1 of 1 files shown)


policycoreutils/setsebool/setsebool.c
```diff
@@ -38,10 +38,7 @@ int main(int argc, char **argv)
 	if (argc < 2)
 		usage();
 
-	if (is_selinux_enabled() <= 0) {
-		fputs("setsebool:  SELinux is disabled.\n", stderr);
-		return 1;
-	}
+	reload = (is_selinux_enabled() > 0);
```
stephensmalley: Alternatively, we could just skip the call to semanage_set_reload() unless -N was specified, as we currently do in semodule and semanage/seobject.py, and then libsemanage would internally check is_selinux_enabled() itself to decide the default for reloading.

policycoreutils/setsebool/setsebool.c
```diff
@@ -130,6 +127,7 @@ static int semanage_set_boolean_list(size_t boolcnt,
 	semanage_bool_key_t *bool_key = NULL;
 	int managed;
 	int result;
+	int enabled = is_selinux_enabled();
```
stephensmalley: Technically, for consistency, this would be int enabled = (is_selinux_enabled() > 0);
However, it is no longer possible for is_selinux_enabled() to return < 0; that was an artifact of when is_selinux_enabled() used to probe for selinuxfs, so alternatively we could just use is_selinux_enabled() alone in both cases.

dustymabe:
thanks @stephensmalley !

PR #130, opened 2019-01-17
libselinux/audit2why.so: Filter out non-python related symbols

stephensmalley:
audit2why.so used to export libsepol.a symbols. We only need Python related
symbols:

- initaudit2why for python 2
- PyInit_audit2why for python3


bachradsusi:
https://travis-ci.org/SELinuxProject/selinux/builds/483832670
https://github.com/SELinuxProject/selinux/commit/de5d5ede603bfad6f12fcd8fe623c8294767ff1c

Issue #137, opened 2019-03-18
restorecond doesn't need pcre


bigon:
Hi,

Any reasons why restorecond is being linked against libpcre?

AFAICS it's not using any symbols from that library

stephensmalley:
Probably unnecessary, indirect dependency via libselinux.

bigon:
What I mean is that it's explicitly liking against it: https://github.com/SELinuxProject/selinux/blob/master/restorecond/Makefile#L22

PR #139, opened 2019-03-21
restorecond: Do not link against libpcre

bigon:
For some reasons, restorecond was explicitly linking against libpcre but
the code is not using any of its symbols

Closes: https://github.com/SELinuxProject/selinux/issues/137


bachradsusi:
Please send the patch for review to the SELinux mailing list at selinux@vger.kernel.org

fishilico:
Merged as https://github.com/SELinuxProject/selinux/commit/42f73af5076411de3c6a5b2bed21a4fd6ae44acd.
Thanks!

PR #144, opened 2019-04-16
checkmodule: allow specifying module policy version

garyttierney:
Currently checkpolicy can produce binary policies for earlier policy versions
to provide support for building policies on one machine and loading/analyzing
them on another machine with an earlier version of the kernel or libsepol,
respectively. However, checkmodule was lacking this capability.
    
This commit adds an identical `-c` flag that can be passed to checkmodule that
will build a modular policy file of the specified version.

Additionally, I updated dismod to display the policy version since there was no way to get at this information without debug logging in the userspace tools (that I know of).

```
> $ cat > test.te <<EOF
module test 1.0;

require {
    type domain;
    type file_type;
    class file { read write };
}

allow domain file_type : file { read write };
EOF
> $ obj/usr/bin/checkmodule -m -M -c 10 -o test.mod test.te
> $ checkpolicy/test/dismod test.mod
Reading policy...
... snip ...
Binary policy module file loaded.
Module name: test
Module version: 1.0
Policy version: 10
```

This patch relates to an issue that was reported by a user in the #selinux Freenode IRC channel who had a later version of libsepol on the machine that was building their policy modules.

### DIFF
--- a/checkpolicy/checkmodule.8
+++ b/checkpolicy/checkmodule.8
@@ -38,7 +38,7 @@ Generate a non-base policy module.
 Enable the MLS/MCS support when checking and compiling the policy module.
 .TP
 .B \-V,\-\-version
- Show policy versions created by this program.  Note that you cannot currently build older versions.
+Show policy versions created by this program.
 .TP
 .B \-o,\-\-output filename
 Write a binary policy module file to the specified filename.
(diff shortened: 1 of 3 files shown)


garyttierney:
I've seen a bunch of PRs come in recently so I'm not sure what the current preferred method of receiving patches is; would it be better to send these changes to the list instead?

doverride:
nit: `type domain;` and `type file_type;` in description are wrong. should be `attribute domain,, file_type;` , not sure why the compiler allows this.

doverride:
I suppose the compiler does not care.

garyttierney:
Oops, you're right -- we're not linking the module against anything, which is why it's allowed.

garyttierney:
Closed by https://github.com/SELinuxProject/selinux/commit/4984a75f3824de88af1dee8635878592ff64155d and https://github.com/SELinuxProject/selinux/commit/ac2a3fb4fedefdd0bcb0fae3e87eb1ea62c3eb2b.

PR #147, opened 2019-05-10
Unify code style to preserve my sanity

UntoSten:
Sorry, this patch contributes very little, but it is essential to preserve my sanity when reading the code.


stephensmalley:
Please submit as a patch to selinux@vger.kernel.org.  If that poses a problem for you, then ask and perhaps someone else can submit it on your behalf.

UntoSten:
Sorry Stephen, I was too hasty in creating these pull requests. I will do as you suggest. Thanks.

PR #148, opened 2019-05-10
Global replace exit(0) with more readable exit(EXIT_SUCCESS)

UntoSten:
Another trivial patch to improve code readability. I did not replace exit(1) with EXIT_FAILURE, because some of the code uses exit(2), exit(3) etc, so preferred to keep the raw integers there.


UntoSten:
Sent this to the mailing list as well.

PR #149, opened 2019-05-10
Check strdup() failure

UntoSten:
Paranoia but should be done


williamcroberts:
I didn't see this hit the mailing list, are you sending it that way or did I miss it?
http://vger.kernel.org/vger-lists.html#selinux

UntoSten:
Sorry, I just sent this to the mailing list...

PR #161, opened 2019-06-14
selinux: Remove legacy local boolean and user code

stephensmalley:
Remove legacy local boolean and user code, and to preserve API/ABI
compatibility the following functions int values should be set to '0'
as they are no longer used:
  selinux_mkload_policy(int preservebools)
  security_set_boolean_list(.... int permanent)
and the following are now no-op:
  security_load_booleans(char *path) - return '0'
  selinux_booleans_path(void) - return NULL
  selinux_users_path(void) - return NULL

Also "SETLOCALDEFS" removed from SELinux config file and code.

Signed-off-by: Richard Haines <richard_c_haines@btinternet.com>


stephensmalley:
Obsoleted by new patch version

PR #162, opened 2019-06-14
libselinux: Use Python distutils to install SELinux python bindings

stephensmalley:
Follow officially documented way how to build C extension modules using
distutils - https://docs.python.org/3.8/extending/building.html#building

Fixes:

- https://bugzilla.redhat.com/show_bug.cgi?id=1715589 - selinux python module
fails to load when it's built using SWIG-4.0:

>>> import selinux
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.7/site-packages/selinux/__init__.py", line 13, in <module>
    from . import _selinux
ImportError: cannot import name '_selinux' from 'selinux' (/usr/lib64/python3.7/site-packages/selinux/__init__.py)
>>>

SWIG-4.0 changed (again?) its behavior so that it uses: from . import _selinux
which looks for _selinux module in the same directory as where __init__.py is -
$(PYLIBDIR)/site-packages/selinux. But _selinux module is installed into
$(PYLIBDIR)/site-packages/ since a9604c30a5e2f ("libselinux: Change the location
of _selinux.so").

On the other hand, the module can't be loaded when it's built using older SWIG.
I'd use a symlink from $(PYLIBDIR)/site-packages/selinux to
$(PYLIBDIR)/site-packages/ in order to fix that

- https://bugzilla.redhat.com/show_bug.cgi?id=1715756 - audit2why python module
fails to build with Python 3.8

- https://bugzilla.redhat.com/show_bug.cgi?id=1709576 - selinux python module
doesn't provide any Python metadata

### DIFF
--- a/libselinux/src/.gitignore
+++ b/libselinux/src/.gitignore
@@ -1,4 +1,4 @@
 selinux.py
-selinuxswig_wrap.c
+selinuxswig_python_wrap.c
 selinuxswig_python_exception.i
 selinuxswig_ruby_wrap.c
--- a/libselinux/src/Makefile
+++ b/libselinux/src/Makefile
@@ -36,7 +36,7 @@ TARGET=libselinux.so
 LIBPC=libselinux.pc
 SWIGIF= selinuxswig_python.i selinuxswig_python_exception.i
 SWIGRUBYIF= selinuxswig_ruby.i
-SWIGCOUT= selinuxswig_wrap.c
+SWIGCOUT= selinuxswig_python_wrap.c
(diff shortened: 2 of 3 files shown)


bachradsusi:
The patch was updated and merged - https://github.com/SELinuxProject/selinux/commit/2efa06857575e4118e91ca250b6b92da68b130d5 - https://lore.kernel.org/selinux/pjd4l3mezyk.fsf@redhat.com/T/#t

PR #163, opened 2019-06-17
libsepol fixes

williamcroberts:
8292fce32195  libsepol: Replace constant with sizeof()
bd396fb2216d libsepol: Check strdup() failures


williamcroberts:
whoops bad PR

PR #173, opened 2019-08-05
Fix many misspellings

fishilico:
Use codespell (https://github.com/codespell-project/codespell) in order to find many common misspellings that are present in English texts.

I am also sending this to the mailing list. I opened this Pull Request because I find it easier to review such changes with GitHub's syntax coloration than with `patch`'s line-based diff format.


fishilico:
Merged as https://github.com/SELinuxProject/selinux/commit/b550c0e20221cca29ac51e6165feec3640baa3a5

Issue #176, opened 2019-09-17
broken setting of log level in libsepol?


mikhailnov:
I wanted to set maximum log level in `semodule` (`libsepol`). Grepped the source code and it seemed me that `lvl` is always `1` and there is no way to change it via cli.

So I had to patch it to get maximum log level.

```diff
diff --git a/libsepol/cil/src/cil_log.c b/libsepol/cil/src/cil_log.c
index b222b155..82dc56f3 100644
--- a/libsepol/cil/src/cil_log.c
+++ b/libsepol/cil/src/cil_log.c
@@ -51,11 +51,11 @@ void cil_set_log_handler(void (*handler)(int lvl, char *msg))
 
 __attribute__ ((format (printf, 2, 0))) void cil_vlog(enum cil_log_level lvl, const char *msg, va_list args)
 {
-	if (cil_log_level >= lvl) {
+	//if (cil_log_level >= lvl) {
 		char buff[MAX_LOG_SIZE];
 		vsnprintf(buff, MAX_LOG_SIZE, msg, args);
 		(*cil_log_handler)(cil_log_level, buff);
-	}
+	//}
 }
 
 __attribute__ ((format (printf, 2, 3))) void cil_log(enum cil_log_level lvl, const char *msg, ...)
```

Please correct me if I'm wrong.

stephensmalley:
cil_set_log_level(), called by secilc or secil2conf with a level that can be incremented by passing one or more -v options?

mikhailnov:
I needed it from `semodule`, not secil

stephensmalley:
Probably semodule should be patched to call cil_set_log_level() with a value based on its existing -v option, which presently only controls its own logging.

PR #182, opened 2019-10-21
policycoreutils: semodule: Enable CIL logging

perfinion:
semodule -v will turn on semodule's own verbose logging but not logging
from CIL. This change makes the verbose flag also set cil's log level.

By default (ie no -v flag), this will enable CIL_ERR, and each -v will
increase the level from there.

Tested with a duplicated fcontext in the policy.
Before this change:
    # semodule -v -B
    Committing changes:
    Problems processing filecon rules
    Failed post db handling
    semodule:  Failed!

After this change:
    # semodule -v -B
    [ ... snip ... ]
    Found conflicting filecon rules
      at /var/lib/selinux/mcs/tmp/modules/400/mycustom/cil:159
      at /var/lib/selinux/mcs/tmp/modules/400/mycustom/cil:158
    Problems processing filecon rules
    Failed post db handling
    semodule:  Failed!

Closes: https://github.com/SELinuxProject/selinux/issues/176
Signed-off-by: Jason Zaman <jason@perfinion.com>

### DIFF
--- a/policycoreutils/semodule/semodule.c
+++ b/policycoreutils/semodule/semodule.c
@@ -22,6 +22,7 @@
 #include <libgen.h>
 #include <limits.h>
 
+#include <sepol/cil/cil.h>
 #include <semanage/modules.h>
 
 enum client_modes {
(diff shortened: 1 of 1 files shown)


perfinion:
Merged

PR #183, opened 2019-10-25
src/Makefile: don't pass bogus -I and -L to python setup.py build_ext

tpetazzoni:
Using $(DESTDIR) during the build does not follow the normal/standard
semantic of DESTDIR: it is normally only needed during the
installation. Therefore, a lot of build systems/environments don't
pass any DESTDIR at build time, which causes setup.py to be called
with -I /usr/include -L /usr/lib, which breaks cross-compilation.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@bootlin.com>


fishilico:
Merged as https://github.com/SELinuxProject/selinux/commit/0271bf4136e3a2bb476919eb9e984418ab95cb75. Thanks!

PR #184, opened 2019-10-25
Build fixes

tpetazzoni:
This pull request contains two build fixes that were needed in the context of packaging SELinux in the Buildroot project. They are sent in a single pull request because even though they address different issues, the second commit modifies the same line of code as the first commit, even if for two completely different reasons.


stephensmalley:
Can you send these via git send-email or equivalent to selinux @ vger.kernel.org?  That's the preferred way for submitting selinux patches.

tpetazzoni:
Ah, yes, sure thing!

PR #185, opened 2019-10-31
restorecond: Fix redundant console log output error

kongbc:
When starting restorecond without any option the following redundant
console log is outputed:

/dev/log 100.0%
/var/volatile/run/syslogd.pid 100.0%
...

This is caused by two global variables of same name r_opts. When
executes r_opts = opts in restore_init(), it originally intends
to assign the address of struct r_opts in "restorecond.c" to the
pointer *r_opts in "restore.c".

However, the address is assigned to the struct r_opts and covers
the value of low eight bytes in it. That causes unexpected value
of member varibale 'nochange' and 'verbose' in struct r_opts, thus
affects value of 'restorecon_flags' and executes unexpected operations
when restorecon the files such as the redundant console log output or
file label nochange.

Signed-off-by: kong baichuan <kongbaichuan@huawei.com>

### DIFF
--- a/restorecond/restore.c
+++ b/restorecond/restore.c
@@ -12,39 +12,36 @@
 char **exclude_list;
 int exclude_count;
 
-struct restore_opts *r_opts;
-
 void restore_init(struct restore_opts *opts)
 {
 	int rc;
 
-	r_opts = opts;
 	struct selinux_opt selinux_opts[] = {
-		{ SELABEL_OPT_VALIDATE, r_opts->selabel_opt_validate },
-		{ SELABEL_OPT_PATH, r_opts->selabel_opt_path },
-		{ SELABEL_OPT_DIGEST, r_opts->selabel_opt_digest }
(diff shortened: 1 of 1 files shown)


stephensmalley:
Please post via git-send-email or equivalent to selinux@vger.kernel.org.

kongbc:
OK, the patch  was sent to selinux@vger.kernel.org.

fishilico:
I have not received the patch, and it did not seem to appear on the mailing list archive (https://lore.kernel.org/selinux/). Did you receive a bounce after sending the patch? Are you subscribed to the mailing list (if not, a moderator might need to unblock your message)?

kongbc:
sorry, I do not subscribe to the mailing list and my message may be blocked. I resend the patch again and it has shown in the email list. please verify it. tks!

stephensmalley:
Merged via ad2208ec220f55877a4d31084be2b4d6413ee082

PR #200, opened 2020-01-23
libsepol,checkpolicy: remove use of hardcoded security class values

stephensmalley:
libsepol carried its own (outdated) copy of flask.h with the generated
security class and initial SID values for use by the policy
compiler and the forked copy of the security server code
leveraged by tools such as audit2why.  Convert libsepol and
checkpolicy entirely to looking up class values from the policy,
remove the SECCLASS_* definitions from its flask.h header, and move
the header with its remaining initial SID definitions private to
libsepol.  While we are here, fix the sepol_compute_sid() logic to
properly support features long since added to the policy and kernel,
although there are no users of it other than checkpolicy -d (debug)
and it is not exported to users of the shared library.  There
are still some residual differences between the kernel logic and
libsepol.


stephensmalley:
@pebenito setools currently depends on sepol/policydb/flask.h for some of these SECCLASS_ definitions, see setools/policyrep/sepol.pxd and fscontext.pxi.  How hard would it be to fix it to look them up instead as in this patch?

pebenito:
It will take a small bit of reworking, but it should be fairly straightforward. It's only used to figure out the class for the genfscons. I guess I can use `symbol.v.sclass` (cython notation--it figures out what is actually a dereference) to look up the class struct with the policydb `class_val_to_struct`, right?

stephensmalley:
There is a similar change in the libsepol/checkpolicy patch to switch from using the #define'd classes to looking them up via hashtab_search(policydb->p_classes.table, "file") or whatever when compiling the genfscon statements. You need to map the string names to the class value in the policy not reverse map it.

stephensmalley:
Aside from removing assumptions from libsepol/checkpolicy itself about hardcoded class values, I also wanted to get this merged because otherwise some people are likely to switch over to using libsepol's flask.h when they see that we have removed libselinux's flask.h instead of correctly migrating to string_to_security_class() and friends.

pebenito:
I have no objections. I have a fix for SETools in my local working tree. It should be up in the next day or two.

stephensmalley:
Ping

stephensmalley:
Merged via f8c110c8a615eb640510eab39640a0957a6ba19c

PR #203, opened 2020-02-12
Fix Ru translation of failsafe context

mikhailnov:
From failsafe_context(5):
"The failsafe_context file allows SELinux-aware applications such as
PAM(8) to obtain a known valid login context for an administrator if
no valid default entries can be found elsewhere."

"Надёжный" means "reliable", "резервный" means "reserve",
the last variant is much closer to what "failsafe" really does.

CCing original translators: @darktemplarbasealt (I forgot Olesya's github account, I think I saw it somewhere)


darktemplarbasealt:
My personal current favorite translation for word "**failsafe**" is "**отказоустойчивый**", although it may vary a lot due to context. It might be a good idea to also consider "**отказобезопасный**" as translation due to context. While these 2 words look quite similar, the meaning is quite different, but it might be not be obvious from quick glance. "**Резервный**" looks good too. **Precise** translation can be tough.

Also, CC @OlesyaGerasimenko

mikhailnov:
Here is a quote from https://github.com/SELinuxProject/selinux/blob/master/libselinux/src/get_context_list.c#L506
```
/* Unable to determine a reachable context list, try to fall back to
   the "failsafe" context to at least permit root login
   for emergency recovery if possible. */
```
"отказоустойчивый" and "отказобезопасный" may make a wrong impression that it is a good idea to use failsafe context configs directly instead of normal ways of configuration. Those words do not show that it is used only if other options are not available, I think "резервный" does show that.

darktemplarbasealt:
Sounds reasonable to me.

bachradsusi:
Thank you for your contribution. In order for your patch to be merged, it needs to be sent to the mailing list selinux@vger.kernel.org (for example by using git format-patch and git send-email)so that  developers can review and test it. If you have trouble doing this, I can forward your patch for you.

mikhailnov:
Really, such a minor fix already approved by the only existing Russian localisator should be emailed for further discussion? Ok, done.

doverride:
On Mon, Feb 17, 2020 at 01:31:19PM -0800, mikhailnov wrote:
> Really, such a minor fix already approved by the only existing Russian localisator should be emailed for further discussion? Ok, done.

Github is not accessible for everyone and others might not want to use Github, but still want have a chance to review and/or comment on the patch. If you dont like the redundancy then probably better to use the maillist for these sorts of things and skip github.

> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/SELinuxProject/selinux/pull/203#issuecomment-587165350

-- 
Key fingerprint = FCD2 3660 5D6B 9D27 7FC6  E0FF DA7E 521F 10F6 4098
https://sks-keyservers.net/pks/lookup?op=get&search=0xDA7E521F10F64098
Dominick Grift

mikhailnov:
> Github is not accessible for everyone and others might not want to use Github, but still want have a chance to review and/or comment on the patch. If you dont like the redundancy then probably better to use the maillist for these sorts of things and skip github.

ok, sounds reasonable. Sorry.

mikhailnov:
As this has been merged manually, I am cloding this PR. Thanks.

PR #209, opened 2020-03-19
Security context t

cgzones:


stephensmalley:
Please post to the mailing list, and the checkpolicy patch should be done separately; it isn't related to this.

PR #229, opened 2020-04-28
libselinux: mount selinuxfs noexec and nosuid

topimiettinen:
Mount selinuxfs with mount flags noexec and nosuid. It's not likely
that this has any effect, but it's visually more pleasing.

Option nodev can't be used because of /sys/fs/selinux/null device,
which is used by Android.


fishilico:
This has been merged as https://github.com/SELinuxProject/selinux/commit/7eaea214a0a5d9e3fb517152ac6162449ed3ef40
Thanks!

PR #232, opened 2020-04-29
sepolicy-gui: fix columns in transitions view

topimiettinen:
Delete an unused column from view "Application Transitions From". The
second column displays names of the executable files instead of
booleans.


fishilico:
This has been merged as https://github.com/SELinuxProject/selinux/commit/db196873ae67ab139c80a9db6b979918e33d42d9
Thanks!

Issue #187, opened 2019-11-11
local test install fails due to missing DESTDIR/usr/lib/python3/dist-packages/selinux 


cgzones:
Building into a local directory fails:

```
make DESTDIR=~/Downloads/selinux_test install install-pywrap

[...]

install -m 644 selinux.py /home/christian/Downloads/selinux_test/usr/lib/python3/dist-packages/selinux/__init__.py
install: cannot create regular file '/home/christian/Downloads/selinux_test/usr/lib/python3/dist-packages/selinux/__init__.py': No such file or directory
make[2]: *** [Makefile:177: install-pywrap] Error 1
make[2]: Leaving directory '/home/christian/Coding/workspaces/selinux_userland/libselinux/src'
make[1]: *** [Makefile:58: install-pywrap] Error 2
make[1]: Leaving directory '/home/christian/Coding/workspaces/selinux_userland/libselinux'
make: *** [Makefile:39: install-pywrap] Error 1
```

Commit 2efa06857575e4118e91ca250b6b92da68b130d5 removed the line ```test -d $(DESTDIR)$(PYTHONLIBDIR)/selinux || install -m 755 -d $(DESTDIR)$(PYTHONLIBDIR)/selinux```.

Maybe reintroduce it, or rework `$(PYTHONLIBDIR)` to contain the python minor version?

stephensmalley:
@bachradsusi Thoughts on this issue?

fishilico:
> Maybe reintroduce it, or rework `$(PYTHONLIBDIR)` to contain the python minor version?

It is kind of strange that `$(PYTHONLIBDIR)` does not contain the minor version (only `/python3/`). Which Linux distribution and which Python interpreter are you using?

In the Makefile, `$(PYTHONLIBDIR)` is computed using a function from module `distutils.sysconfig`. For example, on my development system (Python 3.7.4 on Arch Linux), here is what the used function returns in `python` CLI:
```
>>> from distutils.sysconfig import *
>>> print(get_python_lib(plat_specific=1, prefix='/usr'))
/usr/lib/python3.7/site-packages
```

It would be nice to be possible to add the configuration that causes the issue to Travis-CI configuration in order to detect earlier such a breakage.

cgzones:
Debian sid

```
$ python
Python 2.7.17 (default, Oct 19 2019, 23:36:22) 
[GCC 9.2.1 20191008] on linux2
Type "help", "copyright", "credits" or "license" for more information.
[...]

PR #234, opened 2020-05-01
tree-wide: introduce PYTHON_SETUP_ARGS to customize setup.py calls on …

cgzones:
…Debian

On Debian the `distutils` module is patched, so `get_python_lib()` returns by default `/usr/lib/python3/dist-packages` (no minor version).

But `setuptools` affecting setup.py is not patched to create the library directory at `/usr/lib/python3/dist-packages` by default, rather than a command line argument `--install-layout deb` is added.

Add PYTON_SETUP_ARGS as argument to affected setup.py calls and add a note in the global ReadMe.

See https://www.debian.org/doc/packaging-manuals/python-policy/packaging_tools.html section B.1.

Fixes: https://github.com/SELinuxProject/selinux/issues/187

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ To build and install everything under a private directory, run:
 
     make DESTDIR=~/obj install install-rubywrap install-pywrap
 
+On Debian `PYTHON_SETUP_ARGS=--install-layout=deb` needs to be set when installing the python wrappers in order to create the correct python directory structure.
+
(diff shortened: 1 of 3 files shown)


fishilico:
Merged as https://github.com/SELinuxProject/selinux/commit/7c73601d54427347fda0e5fad134d38aa7ca19c8

PR #238, opened 2020-05-12
mcstransd: fix memory leak in new_context_str

williamcroberts:
The return value of context_new needs to be free with context_free.

Acked-by: William Roberts <william.c.roberts@intel.com>
Signed-off-by: bauen1 <j2468h@gmail.com>


williamcroberts:
Applied by Peter

PR #240, opened 2020-05-15
secilc/docs: fix use of TMPDIR

topimiettinen:
Environment variable TMPDIR may be already set for the user building
and this could be equal to $XDG_RUNTIME_DIR or /tmp which are existing
directories. Then when running 'make clean', there are unintended side
effects:

rm -rf /run/user/1000
rm: cannot remove '/run/user/1000/dconf/user': Permission denied
rm: cannot remove '/run/user/1000/systemd': Permission denied
rm: cannot remove '/run/user/1000/gnupg': Permission denied
rm: cannot remove '/run/user/1000/dbus-1': Is a directory
rm: cannot remove '/run/user/1000/inaccessible': Permission denied
make[1]: *** [Makefile:68: clean] Error 1

Fix by using a different name.

### DIFF
--- a/secilc/docs/Makefile
+++ b/secilc/docs/Makefile
@@ -1,8 +1,8 @@
-CWD ?= $(shell pwd)
-HTMLDIR ?= $(CWD)/html
-PDFDIR ?= $(CWD)/pdf
-TMPDIR ?= $(CWD)/tmp
-TESTDIR ?= $(CWD)/../test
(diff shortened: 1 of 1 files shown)


bachradsusi:
https://github.com/SELinuxProject/selinux/commit/4ad0abd983115bc04c8d8ad12a431cfeabc9555b

PR #250, opened 2020-06-06
fix NULL pointer use in selinux_restorecon_set_sehandle

yanjinjq:
error occur when selinux_restorecon_default_handle return NULL in restorecon_init

### DIFF
--- a/libselinux/src/selinux_restorecon.c
+++ b/libselinux/src/selinux_restorecon.c
@@ -1153,9 +1153,11 @@ void selinux_restorecon_set_sehandle(struct selabel_handle *hndl)
 	size_t num_specfiles, fc_digest_len;
 
 	fc_sehandle = (struct selabel_handle *) hndl;
+	if (!fc_sehandle)
+		return;
 
 	/* Check if digest requested in selabel_open(3), if so use it. */
-	if (fc_sehandle == NULL || selabel_digest(fc_sehandle, &fc_digest, &fc_digest_len,
+	if (selabel_digest(fc_sehandle, &fc_digest, &fc_digest_len,
 				   &specfiles, &num_specfiles) < 0)
 		selabel_no_digest = true;
 	else
--- a/libselinux/src/selinux_restorecon.c
+++ b/libselinux/src/selinux_restorecon.c
@@ -1155,7 +1155,7 @@ void selinux_restorecon_set_sehandle(struct selabel_handle *hndl)
 	fc_sehandle = (struct selabel_handle *) hndl;
 
 	/* Check if digest requested in selabel_open(3), if so use it. */
-	if (selabel_digest(fc_sehandle, &fc_digest, &fc_digest_len,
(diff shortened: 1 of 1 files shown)


bachradsusi:
Hello, thank you for your patch. In order for your patch to be merged, it needs to follow some rules that are written file CONTRIBUTING.md (https://github.com/SELinuxProject/selinux/blob/master/CONTRIBUTING.md). More precisely:

* Please send your patches to the mailing list selinux@vger.kernel.org (for example by using git format-patch and git send-email) in order for the developers to review and test it. If you have trouble doing this, I can forward your patch for you.

 * Please add a "Signed-off-by:" line to the commit message. This means your code meets the Developer's certificate of origin which is written in CONTRIBUTING.md.

libselinux/src/selinux_restorecon.c
```diff
@@ -1155,7 +1155,7 @@ void selinux_restorecon_set_sehandle(struct selabel_handle *hndl)
 	fc_sehandle = (struct selabel_handle *) hndl;
 
 	/* Check if digest requested in selabel_open(3), if so use it. */
-	if (selabel_digest(fc_sehandle, &fc_digest, &fc_digest_len,
+	if (fc_sehandle == NULL || selabel_digest(fc_sehandle, &fc_digest, &fc_digest_len,
```
stephensmalley, line 1158: IMHO, if !hndl we can just return immediately from selinux_restorecon_set_sehandle() without doing anything.  selabel_no_digest is only ever used if fc_sehandle is non-NULL and fc_sehandle is initially NULL.  It just becomes a no-op.  selinux_restorecon() will then return -1 immediately.

yanjinjq: ok , 3ks。Is  it means ok if fix the question like this ? 
~~~
void selinux_restorecon_set_sehandle(struct selabel_handle *hndl)
{
	char **specfiles;
	unsigned char *fc_digest;
	size_t num_specfiles, fc_digest_len;
        
        if (!hndl) {
                return;
        }
	fc_sehandle = (struct selabel_handle *) hndl;

	/* Check if digest requested in selabel_open(3), if so use it. */
	if (selabel_digest(fc_sehandle, &fc_digest, &fc_digest_len,
				   &specfiles, &num_specfiles) < 0)
		selabel_no_digest = true;
	else
		selabel_no_digest = false;
}
~~~

stephensmalley: Probably should return after setting fc_sehandle to hndl just in case anyone uses this to "clear" the fc_sehandle before selabel_close()'ing it.  Also, don't need the { } since it is a single statement body for the if statement.  And I don't see why we have that type cast since it is already that type unless I am missing something.

bachradsusi:
https://github.com/SELinuxProject/selinux/commit/08f5e30177218fae7ce9f5c8d6856690126b2b30

PR #262, opened 2020-08-25
[WIP] Scan with Coverity

cgzones:
This is an attempt to revive Coverity Scan.

Please do not merge but instead push to the branch `coverity_scan` and replace the `secure` token with a value from
https://scan.coverity.com/projects/selinuxproject-selinux/builds/new?tab=travis_ci -> **Example**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,173 +1,47 @@
 # Define the building environment
+os: linux
+dist: bionic
 language: c
 
-matrix:
-  fast_finish: true
+git:
+  depth: 3
 
-compiler:
-  - clang
-  - gcc
 
 env:
-  matrix:
-    # Test the last version of Python and Ruby together, with some linkers
-    - PYVER=python3.8 RUBYLIBVER=2.7
-    - PYVER=python3.8 RUBYLIBVER=2.7 TEST_FLAGS_OVERRIDE=1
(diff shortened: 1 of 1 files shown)


evverx:
@fishilico could you add a secret named `COVERITY_SCAN_TOKEN` to the repository? I'll try to add an action sending data to Coverity daily by analogy with https://github.com/util-linux/util-linux/pull/1491.

evverx:
In the meantime, I think another option would be to set up codeql. I kind of integrated it into my fork in https://github.com/evverx/selinux/pull/1 and it seems to be working
Screenshot 2021-11-16 at 16 14 45

I'll try to polish it a bit and open a PR here

The downside is that all the alerts are hidden in the security tab of repositories, which makes it hard to keep track of reports in one place by external contributors. On the other hand, once codeql ends up in forks they are visible there too. PRs also get analyzed

cgzones:
Travis CI is not used anymore.

evverx:
It should be possible to send data to Coverity Scan using GitHub actions with actions like https://github.com/avahi/avahi/blob/master/.github/workflows/coverity.yml (where https://github.com/avahi/avahi/blob/master/.github/workflows/coverity.sh is used to download Coverity and things like that). It would still be necessary to set up secrets like https://github.com/bus1/dbus-broker/pull/363 though.

PR #264, opened 2020-10-12
libselinux: Add build option to disable X11 backend

Thaodan:

PR #267, opened 2020-10-23
libselinux: Add build option to disable X11 backend

williamcroberts:
Add build option to libselinux to disable the X11 backend without using
ANDROID_HOST.

Signed-off-by: Björn Bidar <bjorn.bidar@jolla.com>


williamcroberts:
Duplicate of #264

PR #277, opened 2021-01-07
libselinux: accept const fromcon in get_context API 

cgzones:
Rework the APIs in <selinux/get_context_list.h> to take a constant
string as from context.

The passed string is not modified currently but not declared const,
which restricting callers (who care about const-correctness).


fishilico:
This was merged as https://github.com/SELinuxProject/selinux/commit/e2dca5df40d110d08b73d5810b88893ba514235e and https://github.com/SELinuxProject/selinux/commit/156dd0de5cad31e7d437c64e11a8aef027f0a691. Thanks!

PR #285, opened 2021-03-19
ci: turn on CIFuzz

evverx:
Now that roughly half the bugs reported by OSS-Fuzz have been
fixed libsepol/cil should be stable enough to get CIFuzz working
more or less reliably. It should help to catch regressions/new bugs
faster.

https://google.github.io/oss-fuzz/getting-started/continuous-integration/

@fishilico could you take a look?

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,39 @@
+---
+name: CIFuzz
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    if: github.repository == 'SELinuxProject/selinux'
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [address, undefined, memory]
+    steps:
+      - name: Build Fuzzers (${{ matrix.sanitizer }})
+        id: build
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+        with:
+          oss-fuzz-project-name: 'selinux'
+          dry-run: false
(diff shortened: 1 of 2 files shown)


bachradsusi:
Please submit all bug reports and patches to <selinux@vger.kernel.org>.

evverx:
Speaking of bug reports, OSS-Fuzz has found 76 bugs so far (40 of which were fixed). I wonder if it would make sense to send those bug reports to the mailing list as well? I could add selinux@vger.kernel.org to https://github.com/google/oss-fuzz/blob/master/projects/selinux/project.yaml.

fishilico:
Hello,

Thanks for your PR. Unfortunately I have been quite busy with other topics this week and I will not have much time to review this properly before 1-2 weeks.

Anyway I am not the only maintainer of SELinux and the other ones still mainly prefer mailing lists (like what is commonly done for Linux kernel development). This is why Petr Lautrbach asked you to send patches to the ML.

> Speaking of bug reports, OSS-Fuzz has found 76 bugs so far (40 of which were fixed). I wonder if it would make sense to send those bug reports to the mailing list as well? I could add selinux@vger.kernel.org to https://github.com/google/oss-fuzz/blob/master/projects/selinux/project.yaml.

As the bug reports are not open (they require authentication with Google accounts), but `selinux@` is an open mailing list, I do not thing it makes much sense to directly cc `selinux@`. Moreover the main developer of the CIL compiler, James Carter, is already in the list of people who receive bug reports and have access to OSS-Fuzz reproducers.

evverx:
I've just sent the patch to the mailing list: https://lore.kernel.org/selinux/20210710120302.74862-1-evvers@ya.ru/T/#t. Closing the PR here.

PR #300, opened 2021-08-03
Improve error message for label file validation

zhangxp1998:


fishilico:
Thanks for the patch. Nevertheless to contribute to this project, patches are required to be sent for review to SELinux mailing list at selinux@vger.kernel.org with a proper `Signed-off-by` tag (see https://github.com/SELinuxProject/selinux/blob/master/CONTRIBUTING.md). Could you please do this?

By the way, your current patch introduces a use-after-free issue: `COMPAT_LOG(..., str_buf)` needs to be called before `free(str_buf);`.

zhangxp1998:
Thanks, I have corrected the user-after-free bug and sent the patch over email.

zhangxp1998:
I got an Acked-by from the mailing list, what do I do next to merge this change?

williamcroberts:
@jwcart2 or @jwcarter  (whats the difference) approved it, so we usually let the person who ack's em merge them. Sometimes they have to generate a PR, but in this case one is open. Assuming it passes CI checks, If for some reason it doesn't get merged in the next couple of days by James ill merge it.

zhangxp1998:
Looks like this got merged, we can close this!

PR #303, opened 2021-08-28
libsepol: fix typo

topimiettinen:
auditdenty -> auditdeny


bachradsusi:
Thanks for the patch, but according to https://github.com/SELinuxProject/selinux/blob/master/CONTRIBUTING.md this needs to be sent for review to SELinux mailing list at selinux@vger.kernel.org with a proper Signed-off-by tag. Could you please do this?

PR #336, opened 2021-12-17
libselinux: make threadsafe for discover_class_cache

purushottamc:
Crash is observed in process dbus-daemon while accessing name
from discover_class_cache structure variable,
discover_class_cache->name variable found NULL
during backtrace analysis.
Add mutex lock for the discover_class_cache to handle multiple
threads for the function which uses discover_class_cache
This avoids variable corruption during parallel access
in the multiple thread environment.

### DIFF
--- a/libselinux/src/stringrep.c
+++ b/libselinux/src/stringrep.c
@@ -19,6 +19,8 @@
 
 #define MAXVECTORS 8*sizeof(access_vector_t)
 
+pthread_mutex_t cache_mutex = PTHREAD_MUTEX_INITIALIZER;
+
 struct discover_class_node {
 	char *name;
 	security_class_t value;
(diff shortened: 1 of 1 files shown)


cgzones:
I think there might still be an issue left:
Consider thread A calling `string_to_security_class()` calling `get_class_cache_entry_name()`, acquiring the mutex, getting the static `discover_class_cache` pointer, searching the required node, releasing the mutex and returning the node pointer back to `string_to_security_class()`.
Now while operating in `string_to_security_class()` on the returned node pointer thread B calls `selinux_flush_class_cache()`, which invalidates all node pointers behind the static `discover_class_cache` and thus thread A will access free'd memory.

Also as by https://github.com/SELinuxProject/selinux/blob/master/CONTRIBUTING.md#contributing-code all patches require a Signed-off-by and needs to be send to selinux@vger.kernel.org.

purushottamc:
Hi @cgzones,
Sent patch to selinux@vger.kernel.org
Did changes as per above review comments but changes are not working as expected with our build system:
https://lore.kernel.org/selinux/20220120073329.15234-1-purushottamchoudhary29@gmail.com/
Could you please have a look into it?

stephensmalley:
Re-opening since this still seems to be an unresolved issue; may try to revisit this.

stephensmalley:
I know it's been a minute since this PR was opened but PTAL at this RFC series, https://lore.kernel.org/selinux/20260612170856.17904-1-stephen.smalley.work@gmail.com/T/#t

stephensmalley:
See https://lore.kernel.org/selinux/20260612191146.20122-1-stephen.smalley.work@gmail.com/T/#t

stephensmalley:
Resolved via the commits referenced above.

PR #349, opened 2022-04-01
sepolgen: Update permission map

bachradsusi:
Use perm_map from SETools project
https://github.com/SELinuxProject/setools/blob/f0ca4667978ec85fe7e9a55c5470d54d88455f96/setools/perm_map


stephensmalley:
Open to re-syncing this but should do so with a current version.

PR #357, opened 2022-06-16
Annotate libselinux functions

cgzones:


cgzones:
systemd build warning:
```
../src/shared/selinux-util.c:558:9: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
        setfscreatecon_raw(NULL);
        ^~~~~~~~~~~~~~~~~~ ~~~~
../src/shared/selinux-util.c:585:9: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
        setsockcreatecon_raw(NULL);
        ^~~~~~~~~~~~~~~~~~~~ ~~~~
```

PR #358, opened 2022-06-22
python/chcat: Don't fail on missing translation files

jsegitz:
Exception handling was not in line with other files, causing chcat to
fail if translation files were not available


bachradsusi:
Thanks for the patch. Please add Signed-off-by:  line to the commit message and send it to selinux@vger.kernel.org for the review - see https://github.com/SELinuxProject/selinux/blob/master/CONTRIBUTING.md#contributing-code

jsegitz:
sorry that one got lost. Thanks, sent it via email

PR #359, opened 2022-06-29
Disjoint Attributes Rule

cgzones:


stephensmalley:
It appeared that this wasn't accepted upstream on the list and has gone stale but feel free to re-open if you choose to revisit it.

PR #365, opened 2022-08-24
Allow user to set SYSCONFDIR for libsemanage

matt-sheets:
This change will allow a user to set the location of their
sysconfdir, defaulted to /etc, if they are installing into
nonstandard locations.


bachradsusi:
This is already merged. Thanks!

PR #383, opened 2023-02-23
setfiles: Add new argument '-y' to set a restorecon like run mode

jsegitz:
This allows to use the text representation of labels to be used for labeling in a restorecon like mode. Useful in e.g. initrd where no policy is loaded


stephensmalley:
Sorry, not sure why this was ignored for so long but if you still care, you should re-base and submit this to selinux@vger.kernel.org via git send-email. If you are unable to send it inline without mangling via git send-email, we can do that part for you.

PR #385, opened 2023-03-30
make avc audit line only have one space

smore-lore:
The kernel-based logs only have one space. So, many scripts parse `avc: denied` and `avc:  denied`. Also, I frequently have developers missing denials because they grep for the single space version locally.


smore-lore:
Fixing the one-space versions instead.

PR #398, opened 2023-06-08
libselinux: Add CPPFLAGS to Makefile

samsam2310:
Add CPPFLAGS to Makefile to allow users change the flags of preprocessor.
We offen use CFLAGS for compiler flags and use CPPFLAGS for preprocessor.
    
Signed-off-by: Chung-Sheng Wu <chungsheng@google.com>


samsam2310:
@vapier PTAL

vapier (approved):

PR #399, opened 2023-06-26
After release the pointer, assign NULL value

gogolovefish:
Signed-off-by: wanhui <wanhui@uniontech.com>


gogolovefish:
Signed-off-by: wanhui <wanhui@uniontech.com>

PR #400, opened 2023-06-26
After release the pointer, assign NULL value

gogolovefish:
Signed-off-by: wanhui <wanhui@uniontech.com>


stephensmalley:
I don't see the necessity of this patch, but if you care, you should re-base and submit it to the selinux@vger.kernel.org mailing list using git send-email.

PR #401, opened 2023-06-29
libselinux: restore: drop the obsolete LSF transitional API.

rmsc:
The preferred way to enable LSF support on 32 bit systems is to define _FILE_OFFSET_BITS=64 when building selinux.

### DIFF
--- a/libselinux/src/selinux_restorecon.c
+++ b/libselinux/src/selinux_restorecon.c
@@ -437,7 +437,7 @@ static int filespec_add(ino_t ino, const char *con, const char *file,
 	file_spec_t *prevfl, *fl;
 	uint32_t h;
 	int ret;
-	struct stat64 sb;
+	struct stat sb;
 
 	__pthread_mutex_lock(&fl_mutex);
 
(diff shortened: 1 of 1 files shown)


rmsc:
Musl has recently deprecated the legacy API. From their Changelog:

"On the API level, the legacy "LFS64" ("large file support") interfaces, which were provided by macros remapping them to their standard names (#define stat64 stat and similar) have been deprecated and are no longer provided under the _GNU_SOURCE feature profile, only under explicit _LARGEFILE64_SOURCE. The latter will also be removed in a future version. Builds broken by this change can be fixed short-term by adding -D_LARGEFILE64_SOURCE to CFLAGS, but should be fixed to use the standard interfaces."

stephensmalley:
Posted to the mailing list here, https://lore.kernel.org/selinux/20260608131614.308870-1-stephen.smalley.work@gmail.com/T/#u

stephensmalley:
Merged via https://github.com/SELinuxProject/selinux/commit/4bdb748b12661b24517d901d957759cf7447303c

PR #405, opened 2023-08-11
Remove ru translation

bachradsusi:
Complete set of patches from https://lore.kernel.org/selinux/20230809163059.97671-1-jwcart2@gmail.com/T/#m50561c6b743400b850a9259b09b3b594c0afe213

Patch 9/12 didn't get to the mailing due to its size


jwcart2:
Thanks, I was going to create a PR myself today.
The patch that was too big wasn't even the biggest one in the series. It was only 70k. I am not sure what happened there.

PR #431, opened 2024-04-04
libselinux: add build time option to drop hard dependency on libpcre2

cgzones:
Currently libselinux links to libpcre2.  The regex library is used in
the file backend of the selabel database for file context path matching.

Some client applications using libselinux might not use that selabel
functionality, but they still require to load libpcre2.  Examples are
dbus-broker and sshd (where openssh only uses the selabel interfaces to
create ~/.ssh with the default context).

Add a build time option, USE_PCRE2_DLSYM, to drop the hard dependency on
libpcre2 and only load it, if actually needed, at runtime via dlopen(3).
Since loading the database for the file backend takes a couple of
milliseconds performance is not a concern.


stephensmalley:
I closed this due to it being stale for 2 years and evidence of upstream discussion on list; if you intend to revive it, feel free to re-open.

PR #439, opened 2024-08-22
libsepol: make reallocarray check more robust

wbx-github:
Some cross-compiler toolchains dislike the usage of -o /dev/null and fail with: ld: final link failed: file truncated

As stdin and stdout is already redirected to /dev/null there is no need for gcc -o /dev/null.

This fixes an Buildroot autobuild failure for libsepol package: http://autobuild.buildroot.net/results/72b/72baaab4658254f4c5b3348298af1d5c891cacd6/

### DIFF
--- a/libsepol/src/Makefile
+++ b/libsepol/src/Makefile
@@ -31,7 +31,7 @@ endif
 
 # check for reallocarray(3) availability
 H := \#
-ifeq (yes,$(shell printf '${H}include <stdlib.h>\nint main(void){return reallocarray(NULL,0,0)==NULL;}' | $(CC) $(CFLAGS) $(LDFLAGS) -x c -o /dev/null - >/dev/null 2>&1 && echo yes))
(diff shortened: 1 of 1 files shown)


cgzones:
It might be binutils bug, similar to the [hppa](https://sourceware.org/bugzilla/show_bug.cgi?id=19526) one fixed [here](https://github.com/bminor/binutils-gdb/commit/6d4b2867f8441c7bf75dfef6b3c95a10241f9709).

stephensmalley:
Submitted to list at https://lore.kernel.org/selinux/CAEjxPJ7BWdm=MkoDGjGn1FwXQb5qzJpphbsPXx5o5LtfuYkTkw@mail.gmail.com/T/#t
but breaks CI - see the thread.

stephensmalley:
I believe this is obsoleted by fixes to ld so that -o /dev/null is in fact always valid.

PR #440, opened 2024-09-01
libsemanage: Mimic GNU basename() API for non-glibc library e.g. musl

listout:
musl only provides POSIX version of basename and it has also removed providing it via string.h header [1] which now results in compile errors with newer compilers e.g. clang-18

[1] https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7


stephensmalley:
I believe this was fixed upstream via a339594da6f027aed5d66ec6798a3d732df235e4

PR #445, opened 2024-09-28
libselinux: Do not clobber errno of the world

crrodriguez:
libselinux clobbers errno of all consumers (systemd, sshd you name it) because its constructors do not properly save and restore errno. The following program demonstrates it.

int main(void)
{
        assert(errno == 0);
	/* Just a test, it doesn't matter already clobbered */
        if(is_selinux_enabled() < 0) {
                perror("not enabled");
                return 1;
        }
}

if any function sets errno, it is not switched back to the original value and standards DO NOT require errno to the set to zero before entering main

### DIFF
--- a/libselinux/src/init.c
+++ b/libselinux/src/init.c
@@ -146,6 +146,7 @@ void set_selinuxmnt(const char *mnt)
 static void init_lib(void) __attribute__ ((constructor));
 static void init_lib(void)
 {
+	SELINUX_PROTECT_ERRNO;
 	selinux_page_size = sysconf(_SC_PAGE_SIZE);
 	init_selinuxmnt();
 #ifndef ANDROID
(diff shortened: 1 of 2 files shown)


cgzones:
Seems fine, just a nit: the identifier `__selinux_saved_errno` is reserved (due to the leading underscore), maybe use `selinux_saved_errno__`?

Also the main way of contribution is to send patches via `git send-email` to selinux@vger.kernel.org.

stephensmalley:
Submitted to list at https://lore.kernel.org/selinux/20260605184558.280873-1-stephen.smalley.work@gmail.com/T/#u

stephensmalley:
Revised and then merged via https://github.com/SELinuxProject/selinux/commit/91145f18fac47a17ac095a187edd78699a58a112

PR #448, opened 2024-10-23
libselinux,libsepol: Honor PREFIX for shared libraries

laomaiweng:
Use `PREFIX` in `SHLIBDIR` just like for the other directories. Without this, local user installs with `PREFIX` set but not `DESTDIR` attempt to write to system locations.


cgzones:
LGTM, however the main way of contribution is to send the patches via `git send-email` to `selinux@vger.kernel.org`.

stephensmalley:
IIRC Absence of $(PREFIX) for SHLIBDIR was intentional for the world where /lib and /usr/lib were different directories with shared libraries going to the former and static libraries to the latter. Kind of moot with usrmerge. In any event, closing this for now.

PR #455, opened 2024-12-17
CI: Fix and migrate the testsuite part to Testing Farm

WOnder93:
Since the current MacOS + Vagrant setup isn't working reliably any more and the testsuite itself has switched its CI to TF, do the same for userspace as well. Also fix a few other CI issues while there.


WOnder93:
This is now merged manually - closing.

PR #464, opened 2025-02-22
Don't inject matchpathcon_filespec_add64() ifdef __x86_64__ (#463, Debian#1098481)

### LINKED ISSUE #463: selinux 3.8 build fails on x32 due to (erroneous?) asserts in matchpathcon.c
selinux 3.8 fails to build on libc-x32 due to a mismatch in the assumptions of `sizeof(unsigned long)`, `sizeof(uint32_t)`, and `sizeof(__ino_t)`, as seen in https://github.com/SELinuxProject/selinux/blob/main/libselinux/src/matchpathcon.c#L270.

Given the hijinks involved in x32 with the 32-bit address space, 64-bit registers, and 64-bit kernel, making `unsigned long` not 32 bits, I'm actually not sure what the right thing to do here is.  Maybe avoid bare `unsigned long` for `mathpathcon_filesppec_add` IAW sourceware's recommendations on x32/glibc at https://www.sourceware.org/glibc/wiki/x32?

While not necessarily relevant here, this does hit at least the debian x32 port: 
```
cc -Wdate-time -D_FORTIFY_SOURCE=2 -D_LARGEFILE64_SOURCE -g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/<<PKGBUILDDIR>>=. -specs=/usr/share/dpkg/pie-compile.specs -fstack-protector-strong -Wformat -Werror=format-security -fcf-protection -fno-semantic-interposition -Wdate-time -D_FORTIFY_SOURCE=2 -Wall -Wextra -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -I../include -D_GNU_SOURCE -DNO_ANDROID_BACKEND -DUSE_PCRE2 -DPCRE2_CODE_UNIT_WIDTH=8 -DHAVE_STRLCPY -DHAVE_REALLOCARRAY  -c -o matchpathcon.o matchpathcon.c
In file included from matchpathcon.c:1:
matchpathcon.c:270:1: error: static assertion failed: "inode size mismatch"
  270 | static_assert(sizeof(unsigned long) == sizeof(__ino_t), "inode size mismatch");
      | ^~~~~~~~~~~~~
make[3]: *** [Makefile:178: matchpathcon.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: Leaving directory '/<<PKGBUILDDIR>>/src'
make[2]: *** [Makefile:54: all] Error 1
make[2]: Leaving directory '/<<PKGBUILDDIR>>'
make[1]: *** [debian/rules:56: override_dh_auto_build] Error 2
make[1]: Leaving directory '/<<PKGBUILDDIR>>'
make: *** [debian/rules:50: binary-arch] Error 2
```

I suspect the 'right' thing to do is actually test on what x86_64-linux-gnux32/sys/types.h ends up declaring, but I don't know if that's sufficient for selinux.  Hacking the `static_assert`s to do basically that seems to work, but I'm not in a spot where I can test that properly right now.

nabijaczleweli:
As the code notes, it wants to add an
```c
/* ABI backwards-compatible shim for non-LFS 32-bit systems */
```
it tries to detect these with
```c
#if defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64 && __BITS_PER_LONG < 64
```
which is correct with the added precondition that the `ino_t` *without*
`-D_FILE_OFFSET_BITS=64` *was actually* u32 (i.e. it conflates *all* ILP32 systems into being non-LFS).

This is not the case on x32, for example, which is LFS; thus, the
```c
static_assert(sizeof(unsigned long) == sizeof(__ino_t), "inode size mismatch");
```
assertion fails (`__ino_t` is the "kernel ino_t" type, which generally corresponds to the kernel's ulong, which is u64 on x32).

The correct spelling of the test for this is
```c
#if ... && sizeof(__ino_t) == 4
```
but this is not statically solvable with the preprocessor.

Thus, we need to explicitly special-case this. `__x86_64__` indicates one of two ABIs (LP64 (amd64) or ILP32 (x32)), both of which have ino_t=u64, and is the macro used for defining `__INO_T_TYPE` in the system headers, so it's the best fit here.

Fixes: commit 9395cc0322 ("Always build for LFS mode on 32-bit archs.")
Closes: #463
Closes: Debian#1098481
Cc: @kainz

---

Insert -I../../libselinux/include and -L../../libselinux/src into subprograms where needed

These subprograms (sublibraries) use plain
```
#include <selinux/selinux.h>
cc ... -lselinux ...
```
&c. which includes/links to the system libselinux.

Naturally, this doesn't work if you don't have one.

These all came out of a plain build on x32.

### DIFF
--- a/libselinux/include/selinux/selinux.h
+++ b/libselinux/include/selinux/selinux.h
@@ -537,7 +537,7 @@ extern int matchpathcon_index(const char *path,
    with the same inode (e.g. due to multiple hard links).  If so, then
    use the latter of the two specifications based on their order in the 
    file contexts configuration.  Return the used specification index. */
-#if defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64 && __BITS_PER_LONG < 64
+#if defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64 && !defined(__INO_T_MATCHES_INO64_T)
 #define matchpathcon_filespec_add matchpathcon_filespec_add64
 #endif
 extern int matchpathcon_filespec_add(ino_t ino, int specind, const char *file);
--- a/libselinux/src/matchpathcon.c
+++ b/libselinux/src/matchpathcon.c
@@ -261,7 +261,7 @@ int matchpathcon_filespec_add(ino_t ino, int specind, const char *file)
 	return -1;
 }
 
-#if defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64 && __BITS_PER_LONG < 64
+#if (defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64) && !defined(__INO_T_MATCHES_INO64_T)
 /* alias defined in the public header but we undefine it here */
 #undef matchpathcon_filespec_add
 
@@ -280,9 +280,13 @@ int matchpathcon_filespec_add(unsigned long ino, int specind,
 {
 	return matchpathcon_filespec_add64(ino, specind, file);
 }
+#elif (defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64) || defined(__INO_T_MATCHES_INO64_T)
+
+static_assert(sizeof(uint64_t) == sizeof(ino_t), "inode size mismatch");
+
 #else
 
-static_assert(sizeof(unsigned long) == sizeof(ino_t), "inode size mismatch");
+static_assert(sizeof(uint32_t) == sizeof(ino_t), "inode size mismatch");
 
 #endif
 
--- a/libsemanage/src/Makefile
+++ b/libsemanage/src/Makefile
@@ -60,7 +60,7 @@ CFLAGS ?= -Werror -Wall -W -Wundef -Wshadow -Wmissing-noreturn -Wmissing-format-
 SWIG_CFLAGS += -Wno-error -Wno-unused-but-set-variable -Wno-unused-variable -Wno-shadow \
 		-Wno-unused-parameter -Wno-missing-prototypes
 
-override CFLAGS += -I../include -D_GNU_SOURCE
+override CFLAGS += -I../include -I../../libselinux/include -D_GNU_SOURCE
 RANLIB ?= ranlib
 
 SWIG = swig -Wall -python -o $(SWIGCOUT) -outdir ./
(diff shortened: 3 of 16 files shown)


bachradsusi:
Hello, thanks for the patch!

In order to get the patch reviewed and accepted, it's necessary to send it to [selinux@vger.kernel.org](mailto:selinux@vger.kernel.org) mailing list. The best way is to use git send-email command, e.g.:

     git send-email --from='Your Name <your@email.address>' --to=selinux@vger.kernel.org --smtp-server=my.smtp.server --confirm=auto -1

See https://github.com/SELinuxProject/selinux/blob/main/CONTRIBUTING.md

nabijaczleweli:
https://lore.kernel.org/selinux/jievtm5yb2pqqrn5p2idny7iipfoxnwn3gx5xtw6ycq6qqfgh5@tarta.nabijaczleweli.xyz/t/#u
https://lore.kernel.org/selinux/kwnnb3dirxubncipnklpimiv34bmetaxchzvzkoztsw62kx7zd@tarta.nabijaczleweli.xyz/t/#u

mildsunrise:
@nabijaczleweli thank you for the patch, I was about to submit a very similar one for RV32

mildsunrise:
The definition is using `__ino_t` which is, as far as I can see, glibc specific. If using glibc-specific macros is acceptable, then the right criteria would be:

```diff
-#if defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64 && __BITS_PER_LONG < 64
+#if defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64 && __TIMESIZE < 64
```

nabijaczleweli:
I think you nailed it actually yeah

nabijaczleweli:
https://lore.kernel.org/selinux/cpbjftpok3lt54s6t7qmhzwbhl47vsz3efrxs45t5os7ztp3zg@tarta.nabijaczleweli.xyz/t/#u

mildsunrise:
perfect, thank you :)

mildsunrise:
hi sorry to bother you again, just wanted to note that (if I'm not mistaken) this patch breaks builds with `USE_LFS=n` on non-LFS 32-bit systems. I think nobody should be building with `USE_LFS=n` but since it's just a default, wanted to raise it just in case.

of course, the previous assertion

~~~ c
static_assert(sizeof(unsigned long) == sizeof(ino_t), "inode size mismatch");
~~~

was also wrong... I'd suggest a third clause, e.g.:

~~~ c
#if (defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64) && !defined(__INO_T_MATCHES_INO64_T)
/*...*/
#elif (defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64) || !defined(__INO_T_MATCHES_INO64_T)

static_assert(sizeof(uint64_t) == sizeof(ino_t), "inode size mismatch");

#else

static_assert(sizeof(uint32_t) == sizeof(ino_t), "inode size mismatch");

#endif
~~~

mildsunrise:
It seems that before this patch, libselinux didn't compile with musl for 32-bit platforms, and this patch further broke the build for 64-bit platforms too. @alyssais has posted an [improved](https://lore.kernel.org/selinux/20250426151356.7116-2-hi@alyssa.is/) version of this patch which should fix the build for both 32-bit and 64-bit platforms :)

PR #469, opened 2025-04-11
libsemanage: add restorecon config option

RossComputerGuy:
This flag allows for enabling or disabling automatic restorecon that semodule invokes. By default, we have it enabled to produce the same behavior as before. On NixOS, we need this as we're "baking" the module installation into a squashfs image and we cannot run restorecon inside the builder.

https://lore.kernel.org/selinux/20250411183532.42053-1-tristan.ross@midstall.com/T/#u


RossComputerGuy:
https://github.com/SELinuxProject/selinux/commit/2647cc0fdca326b81ee3c08718cbe19b7866b53a

PR #472, opened 2025-04-24
libsepol,checkpolicy: introduce unconfined as a first class concept t…

stephensmalley:
…o SELinux

Introduce unconfined as a first class concept to SELinux. Previously, SELinux had no such concept and unconfined domains were solely defined by policy.

Advantages of elevating to a first class concept:
1. Explicit allow rules for unconfined domains can be omitted, shrinking the policy.
2. The kernel can leverage this knowledge to further optimize checking on unconfined processes.

Disadvantages:
1. Yet another juicy target for exploit writers, similar to permissive types.
2. Requires a new policy version/format and bitmap.


libsepol/src/policydb.c
```diff
@@ -1520,7 +1542,7 @@ void policydb_destroy(policydb_t * p)
 
 	ebitmap_destroy(&p->policycaps);
 
-	ebitmap_destroy(&p->permissive_map);
+	ebitmap_destroy(&p->unconfined_map);
```
cgzones, line 1545: this looks wrong

cgzones:
Maybe add a statement to *https://github.com/SELinuxProject/selinux/blob/main/checkpolicy/tests/policy_allonce.conf* for a trivial round-trip test?

PR #473, opened 2025-04-25
libsepol,checkpolicy: introduce neveraudit types

stephensmalley:
Introduce neveraudit types i.e. types that should never trigger audit messages. This allows the AVC to skip all audit-related processing for such types. Note that neveraudit differs from dontaudit not only wrt being applied for all checks with a given source type but also in that it disables all auditing, not just permission denials.

When a type is both a permissive type and a neveraudit type, the security server can short-circuit the security_compute_av() logic, allowing all permissions and not auditing any permissions.


stephensmalley:
Merged via https://github.com/SELinuxProject/selinux/commit/45c5036a691eaae256e36440f0e8858697cf236e

PR #490, opened 2025-08-28
setfiles: Add -A option to disable SELINUX_RESTORECON_ADD_ASSOC

rwmjones:
SELINUX_RESTORECON_ADD_ASSOC tracks conflicts between inodes that have differing contexts.  However doing this involves building a large internal hashtable that stores the full path of every file examined by setfiles.  For filesystems that have very large numbers of files or long pathnames, this uses a lot of memory, which makes SELinux relabelling in constrained memory environments infeasible.

This adds a new setfiles -A option that disables this tracking.

For example, using setfiles to relabel a filesystem with 15 million files took 3.7GB of RAM.  Using this option, the same filesystem can be relabelled in 121MB (albeit with no warnings or errors possible for conflicting labels, but for our use case we don't care about that.)

Fixes: https://issues.redhat.com/browse/RHEL-111505


rwmjones:
Usage should be updated now.

Issue #506, opened 2026-03-09
Compilation fails with glibc 2.43


dschepler:
In a build of selinux version 3.10 on my system where I recently upgraded to glibc 2.43, I get:
```
cc  -O -Wall -W -Wundef -Wformat-y2k -Wformat-security -Winit-self -Wmissing-include-dirs -Wunused -Wunknown-pragmas -Wstrict-aliasing -Wshadow -Wpointer-arith -Wbad-function-cast -Wcast-align -Wwrite-strings -Waggregate-return -Wstrict-prototypes -Wold-style-definition -Wmissing-prototypes -Wmissing-declarations -Wmissing-noreturn -Wmissing-format-attribute -Wredundant-decls -Wnested-externs -Winline -Winvalid-pch -Wvolatile-register-var -Wdisabled-optimization -Wbuiltin-macro-redefined -Wattributes -Wmultichar -Wdeprecated-declarations -Wdiv-by-zero -Wdouble-promotion -Wendif-labels -Wextra -Wformat-extra-args -Wformat-zero-length -Wformat=2 -Wmultichar -Woverflow -Wpointer-to-int-cast -Wpragmas -Wframe-larger-than=32768 -fstack-protector-all --param=ssp-buffer-size=4 -fexceptions -fasynchronous-unwind-tables -fdiagnostics-show-option -Werror -Wno-aggregate-return -Wlogical-op -Wpacked-bitfield-compat -Wsync-nand -Wcoverage-mismatch -Wcpp -Wformat-contains-nul -Wnormalized=nfc -Wsuggest-attribute=const -Wsuggest-attribute=noreturn -Wsuggest-attribute=pure -Wtrampolines -Wjump-misses-init -Wno-suggest-attribute=pure -Wno-suggest-attribute=const -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -Wstrict-overflow=5 -fno-semantic-interposition -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -I../include -D_GNU_SOURCE -DNO_ANDROID_BACKEND -DUSE_PCRE2 -DPCRE2_CODE_UNIT_WIDTH=8 -DHAVE_STRLCPY -DHAVE_REALLOCARRAY  -c -o selinux_config.o selinux_config.c
selinux_config.c: In function ‘selinux_set_policy_root’:
[...]

PR #507, opened 2026-03-12
policycoreutils/secon: fix discarded-qualifiers warning with glibc 2.43

dustinkirkland:
## Problem

In `my_getXcon_raw()`, `ptr` is declared as `const char *` but is
assigned from `fgets()`, which returns `char *`. With glibc 2.43,
`strchr(const char *, int)` now correctly returns `const char *`
(matching the constness of its input), so the subsequent assignment
to `char *tmp` triggers a build failure:

```
secon.c: In function 'my_getXcon_raw':
secon.c:365:18: error: initialization discards 'const' qualifier from
pointer target type [-Werror=discarded-qualifiers]
  365 |         char *tmp = strchr(ptr, '\n');
      |                     ^~~~~~
```

This breaks builds on any system with glibc 2.43+ and gcc with
`-Werror` (which the SELinux Makefile uses).

## Fix

The `const` on `ptr` was always incorrect:

- `fgets()` returns `char *` (a mutable pointer into `buf`, which is
  `char[4096]`)
- The result of `strchr()` through `ptr` is immediately used mutably:
  `*tmp = 0`

Remove the erroneous `const` qualifier from `ptr`.

## Relation to issue #506

This is the same class of bug reported in #506
(`libselinux/src/selinux_config.c:284`). That issue covers
`strrchr()` in `selinux_set_policy_root()`; this patch covers
`strchr()` in `my_getXcon_raw()` in `policycoreutils/secon/secon.c`.

Both instances have the same root cause: a local variable was
unnecessarily `const`-qualified, which glibc 2.43's strengthened
`strchr`/`strrchr` const-propagation now correctly rejects.

## Testing

Verified by building `policycoreutils` version 3.10 against glibc 2.43
on a Wolfi-based system (Chainguard) where this was a build failure.

Signed-off-by: Dustin Kirkland <dustin.kirkland@chainguard.dev>

### DIFF
--- a/policycoreutils/secon/secon.c
+++ b/policycoreutils/secon/secon.c
@@ -348,7 +348,7 @@ static int my_getXcon_raw(pid_t pid, char  **con, const char *val)
 {
 	char buf[4096];
 	FILE *fp = NULL;
-	const char *ptr = NULL;
+	char *ptr = NULL;
 
 	snprintf(buf, sizeof(buf), "%s/%ld/attr/%s", "/proc", (long int)pid,
 		 val);


bachradsusi:
https://github.com/SELinuxProject/selinux/commit/a33609b92e8a27ef721d17ee1dbef44967802820

PR #511, opened 2026-04-11
expose alias for stat64 and lstat64 on musl

### LINKED ISSUE #477: Musl build failure because of -Wimplicit-function-declaration
When trying to compile it on a Raspberry Pi 4B with musl I get:

```
clang  -O3 -pipe -flto -mcpu=cortex-a72+crc -mtune=cortex-a72 -fvectorize -fomit-frame-pointer -fstack-protector-strong -D_FORTIFY_SOURCE=2 -ftrivial-auto-var-init=zero -fsanitize-trap=signed-integer-overflow,unsigned-integer-overflow -fstack-clash-protection -fPIE -fPIC -fno-semantic-interposition -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -I../include -D_GNU_SOURCE -DNO_ANDROID_BACKEND -DUSE_PCRE2 -DPCRE2_CODE_UNIT_WIDTH=8 -DHAVE_STRLCPY -DHAVE_REALLOCARRAY  -c -o sestatus.o sestatus.c
selinux_restorecon.c:441:16: error: variable has incomplete type 'struct stat64'
  441 |         struct stat64 sb;
      |                       ^
selinux_restorecon.c:441:9: note: forward declaration of 'struct stat64'
  441 |         struct stat64 sb;
      |                ^
selinux_restorecon.c:455:10: error: call to undeclared function 'lstat64'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  455 |                         ret = lstat64(fl->file, &sb);
      |                               ^
selinux_restorecon.c:455:10: note: did you mean 'lstat'?
/usr/include/sys/stat.h:82:5: note: 'lstat' declared here
   82 | int lstat(const char *__restrict, struct stat *__restrict);
      |     ^
clang  -O3 -pipe -flto -mcpu=cortex-a72+crc -mtune=cortex-a72 -fvectorize -fomit-frame-pointer -fstack-protector-strong -D_FORTIFY_SOURCE=2 -ftrivial-auto-var-init=zero -fsanitize-trap=signed-integer-overflow,unsigned-integer-overflow -fstack-clash-protection -fPIE -fPIC -fno-semantic-interposition -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -I../include -D_GNU_SOURCE -DNO_ANDROID_BACKEND -DUSE_PCRE2 -DPCRE2_CODE_UNIT_WIDTH=8 -DHAVE_STRLCPY -DHAVE_REALLOCARRAY  -c -o setenforce.o setenforce.c
2 errors generated.
make[1]: *** [Makefile:178: selinux_restorecon.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/var/tmp/portage/sys-libs/libselinux-3.8.1/work/libselinux-3.8.1-.arm64/src'
make: *** [Makefile:54: all] Error 1
```

peppergrayxyz:
provide alias for `stat64` and `lstat64` for use in `filespec_add`:

https://github.com/SELinuxProject/selinux/blob/6be1ec3792c11040fd7a3ecb1135e54418eb0d57/libselinux/src/selinux_restorecon.c#L440-L446
https://github.com/SELinuxProject/selinux/blob/6be1ec3792c11040fd7a3ecb1135e54418eb0d57/libselinux/src/selinux_restorecon.c#L456-L460

fix: #477

### DIFF
--- a/libselinux/src/selinux_restorecon.c
+++ b/libselinux/src/selinux_restorecon.c
@@ -19,7 +19,6 @@
 #include <limits.h>
 #include <stdint.h>
 #include <sys/types.h>
-#include <sys/stat.h>
 #include <sys/xattr.h>
 #include <sys/vfs.h>
 #include <sys/statvfs.h>
@@ -28,6 +27,10 @@
 #include <libgen.h>
 #include <syslog.h>
 #include <assert.h>
+
+// expose alias for stat64 and lstat64 on musl
+#define _LARGEFILE64_SOURCE 1
+#include <sys/stat.h>
 
 #include <selinux/selinux.h>
 #include <selinux/context.h>


WavyEbuilder:

stephensmalley:
Defining _LARGEFILE64_SOURCE must occur before any system headers are
included, including indirectly included ones, to guarantee consistent
symbol exposure.
In the patch, the macro is added after several system headers
(<limits.h>, <stdint.h>, <sys/types.h>, <sys/xattr.h>, etc.).
If any of these headers conditionally depend on large-file
feature-test macros, this may not produce the intended outcome and can
cause subtle ABI mismatches.

Recommendation:

Move _LARGEFILE64_SOURCE into the global feature macros block at the
top of the file, before any includes.
Alternatively, add it to compiler flags where appropriate
(CFLAGS+=-D_LARGEFILE64_SOURCE), which is what glibc’s docs recommend
for feature-test macros.

Portability implications

_LARGEFILE64_SOURCE is musl-specific nowadays; glibc largely ignores
it because 64‑bit off_t is the default on modern architectures.
If this file is built on systems like Android, BSD, or older glibc,
feature-test macros may have different effects.

Recommendation:

Before adopting this patch as-is, consider guarding it: #ifdef musl
#define _LARGEFILE64_SOURCE 1 #endif

This ensures the macro is only used where required.

Commit message / context

The commit message “fix: 477” is minimal. For upstream SELinux
projects, a more complete explanation of the musl build issue is
usually expected.
The email explanation provides context, but most of it should be in
the commit message body rather than outside it.

thesamesam:
Yes, it must be set before any includes, and consistently so. Mismatches between TUs is problematic.

>  _LARGEFILE64_SOURCE is musl-specific nowadays; glibc largely ignores it because 64‑bit off_t is the default on modern architectures.

This isn't right, but neither is the PR as it stands.

 `_LARGEFILE64_SOURCE` gates some dual APIs that glibc will expose if it's set. musl used to unconditionally expose these functions and types, it now gates it behind a macro, and it plans to remove them entirely. The APIs in question never really took off.

> glibc largely ignores it because 64‑bit off_t is the default on modern architectures.

I don't think it ever really ignores it. Not on 32-bit arches, though, unless you pass `_FILE_OFFSET_BITS` with glibc.

> Before adopting this patch as-is, consider guarding it: #ifdef musl
> #define _LARGEFILE64_SOURCE 1 #endif

This won't work:
a) musl famously doesn't define any macro to detect it;
b) the macro isn't musl-specific anyway: if you're using those APIs, define the macro for it consistently;
c) musl plans to remove support for these entirely.

The right fix is to ensure `-D_FILE_OFFSET_BITS=64` is passed everywhere (which makes it safe to..) then always use the regular APIs. I wrote a bit more about it at https://wiki.gentoo.org/wiki/Musl_porting_notes#error:_LFS64_interfaces_.28.2A64_undeclared_here.2C_ex._pread64.29

I would just fix it myself but I was hoping a musl user (who is obviously more motivated) would handle it.

stephensmalley:
libselinux/Makefile has this:
`USE_LFS ?= y
ifeq ($(USE_LFS),y)
        LFS_CFLAGS := -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64
endif
export LFS_CFLAGS
`

stephensmalley:
So can we just unconditionally replace struct stat64 w/ struct stat and lstat64() with lstat() in src/selinux_restorecon.c and call it done?

stephensmalley:
This seems to be what #401 does

stephensmalley:
I took #401 to the mailing list via https://lore.kernel.org/selinux/20260608131614.308870-1-stephen.smalley.work@gmail.com/T/#u

thesamesam:
> So can we just unconditionally replace struct stat64 w/ struct stat and lstat64() with lstat() in src/selinux_restorecon.c and call it done?

Yes indeed.

stephensmalley:
Fixed via https://github.com/SELinuxProject/selinux/commit/4bdb748b12661b24517d901d957759cf7447303c

PR #513, opened 2026-04-11
define matchpathcon_filespec_add/add64

### LINKED ISSUE #512: 'matchpathcon_filespec_add64': symbol not defined
build on musl/llvm fails with:

```
ld.lld: error: version script assignment of 'LIBSELINUX_3.8' to symbol 'matchpathcon_filespec_add64' failed: symbol not defined
cc: error: linker command failed with exit code 1 (use -v to see invocation)
``` 

see also: https://github.com/SELinuxProject/selinux/issues/477#issuecomment-3025566009
downstream bug: https://bugs.gentoo.org/972534

peppergrayxyz:
`matchpathcon_filespec_add` and `matchpathcon_filespec_add64` are exported in `libselinux.map`, so both need to be linker-visible symbols.

https://github.com/SELinuxProject/selinux/blob/6be1ec3792c11040fd7a3ecb1135e54418eb0d57/libselinux/src/libselinux.map#L85
https://github.com/SELinuxProject/selinux/blob/6be1ec3792c11040fd7a3ecb1135e54418eb0d57/libselinux/src/libselinux.map#L258

fix: #512

### DIFF
--- a/libselinux/include/selinux/selinux.h
+++ b/libselinux/include/selinux/selinux.h
@@ -537,10 +537,8 @@ extern int matchpathcon_index(const char *path,
    with the same inode (e.g. due to multiple hard links).  If so, then
    use the latter of the two specifications based on their order in the 
    file contexts configuration.  Return the used specification index. */
-#if defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64 && defined(__INO64_T_TYPE) && !defined(__INO_T_MATCHES_INO64_T)
-#define matchpathcon_filespec_add matchpathcon_filespec_add64
-#endif
 extern int matchpathcon_filespec_add(ino_t ino, int specind, const char *file);
+extern int matchpathcon_filespec_add64(ino_t ino, int specind, const char *file);
 
 /* Destroy any inode associations that have been added, e.g. to restart
    for a new filesystem. */
--- a/libselinux/src/matchpathcon.c
+++ b/libselinux/src/matchpathcon.c
@@ -261,25 +261,20 @@ int matchpathcon_filespec_add(ino_t ino, int specind, const char *file)
 	return -1;
 }
 
+int matchpathcon_filespec_add64(unsigned long ino, int specind,
+                              const char *file)
+{
+	return matchpathcon_filespec_add(ino, specind, file);
+}
+
 #if (defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64) && defined(__INO64_T_TYPE) && !defined(__INO_T_MATCHES_INO64_T)
-/* alias defined in the public header but we undefine it here */
-#undef matchpathcon_filespec_add
 
 /* ABI backwards-compatible shim for non-LFS 32-bit systems */
-
 static_assert(sizeof(unsigned long) == sizeof(__ino_t), "inode size mismatch");
 static_assert(sizeof(unsigned long) == sizeof(uint32_t), "inode size mismatch");
 static_assert(sizeof(ino_t) == sizeof(ino64_t), "inode size mismatch");
 static_assert(sizeof(ino64_t) == sizeof(uint64_t), "inode size mismatch");
 
-extern int matchpathcon_filespec_add(unsigned long ino, int specind,
-                                     const char *file);
-
-int matchpathcon_filespec_add(unsigned long ino, int specind,
-                              const char *file)
-{
-	return matchpathcon_filespec_add64(ino, specind, file);
-}
 #elif (defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS == 64) || defined(__INO_T_MATCHES_INO64_T)
 
 static_assert(sizeof(uint64_t) == sizeof(ino_t), "inode size mismatch");


WavyEbuilder:

bachradsusi:
https://lore.kernel.org/selinux/CAP+JOzRSBYSjUQBjv8+Nt6KrJG-H=g3azyVD9dnT1GOEQyxAdA@mail.gmail.com/

stephensmalley:
Please resolve conflicts. If you are unable to post to the mailing list, we can do that part for you.

stephensmalley:
I apologize but I am not sure if this is the correct fix; it seems to undo a part of  9395cc03226a0e1a220 ("Always build for LFS mode on 32-bit archs.") and I am unclear if the intent of that commit is preserved. It could be, but I am uncertain. I replied on-list with the author of that commit cc'd but unfortunately that bounced so I guess he has moved on. Would appreciate any clarification you or others can provide.

stephensmalley:
On Fedora we have the following:
> $ nm /usr/lib/debug/usr/lib/libselinux.so.1-3.10-1.fc44.i386.debug | grep matchpathcon_filespec_add
> 00015610 T matchpathcon_filespec_add
> 00015330` T matchpathcon_filespec_add64
> 00015330 t matchpathcon_filespec_add64.localalias

> $ nm /usr/lib/debug/usr/lib64/libselinux.so.1-3.10-1.fc44.x86_64.debug | grep matchpathcon_filespec_add
> 0000000000015690 T matchpathcon_filespec_add

Hence, the _add64 symbol only exists on 32-bit which IIUC based on the commit that introduced it, was intentional.
Your patch changes that and makes it unconditional.

stephensmalley:
Per Claude:

The actual issue is that lld (LLVM's linker) is stricter than GNU ld
about version scripts: it treats a version script entry for a symbol
that doesn't exist in the compiled output as a hard error, rather than
silently ignoring it. GNU ld lets you list symbols in a version script
that aren't present; lld does not.
On musl + llvm, matchpathcon_filespec_add64 is only compiled into the
.so when _FILE_OFFSET_BITS == 64 && __BITS_PER_LONG < 64 — i.e., only
on 32-bit targets. On a 64-bit musl build with lld, the symbol doesn't
exist in the object code, but it's unconditionally listed in
libselinux.map, so lld hard-errors out.
The fix needs to be one of:

1. Conditionalize the map entry — not possible directly since version
scripts don't support #ifdef, but the map file can be preprocessed
(generated via cpp or sed from a .[map.in](http://map.in/) template) so the stanza is
only emitted on 32-bit targets.

2. Use --undefined-version to suppress the lld error — a linker flag,
not a source fix, so it's a workaround rather than a proper solution.

3. Always compile matchpathcon_filespec_add64 regardless of
architecture, making the symbol unconditionally present and satisfying
both GNU ld and lld. This is what the referenced issue #477 comment
suggests — the function is harmless on 64-bit where ino_t is already
64 bits.

Option 3 is probably the cleanest upstream fix: remove the #if guard
around the matchpathcon_filespec_add64 definition in matchpathcon.c so
it's always compiled, matching what the map file unconditionally
declares.

That suggests that your patch is in fact the correct fix.

stephensmalley:
Re-based and posted to list at:
https://lore.kernel.org/selinux/20260610133109.6726-1-stephen.smalley.work@gmail.com/T/#u

stephensmalley:
Sorry, this still seems wrong to me. The _add64 interface should NOT devolve to the same as the _add interface.

stephensmalley:
(when dealing with 32-bit ino_t)

stephensmalley:
A possible fix can be found at https://lore.kernel.org/selinux/20260610193945.10035-1-stephen.smalley.work@gmail.com/T/#u

eli-schwartz:
I would personally suggest that one simply doesn't use a broken linker that intentionally has buggy behavior.

If one must use lld, "Claude" is (stochastically) wrong (as it always is) in simultaneously claiming that it's a "hard error" and "is not allowed" while also claiming that there's an option to suppress the error.

"Claude" is also (stochastically) wrong in claiming that it's "a linker flag, not a source fix, so it's a workaround rather than a proper solution". Since when is a linker flag "not a proper solution because linker flags aren't source code"? What does that even mean? 

It is perfectly valid to pass linker flags in a project build system. That flag exists in both GNU `ld.bfd` and LLVM `ld.lld` along with its inverse, `--no-undefined-version`.

The issue is that GNU ld recognizes that it is a common use case to have a linker version script that is useful for multiple build configurations -- or multiple operating systems where one symbol is only present on say, Solaris. People have been doing it for decades, since before LLVM existed. GNU ld defaults to --undefined-version and lets you pass an option to toggle the default if you wish to.

LLVM lld thinks that people should write multiple version scripts (or configure their version scripts using sed, perhaps) in order to "catch bugs where people forgot to perform dead code elimination and delete unused lines". (Why is this a useful thing to error by default, on?) So they default to --no-undefined-version. It is an *opinion*, and people therefore end up applying what is essentially a `-Werror=unused` (that should be a project option) as a side effect of switching to a linker that claims to be "faster than GNU".

It is completely legitimate for a project to disable a "coding style" warning that is incorrect for their use cases. Selinux is using that symbol, and LLVM is erroneously assuming that the symbol was from a previous version and is now removed, just because it happens to not be used in this specific build configuration.

So: just disable the coding style warning. ;) Pass `-Wl,--undefined-version`.

eli-schwartz:
It seems quite wrong to me to grow the symbol table with a symbol that is not intended to ever be used, just to work around an opinionated default warning in lld. If the symbol isn't ever supposed to be used then why advertise it? Or shouldn't the function body be a `trap("how did you get here");` instead?

If there's an unrelated reason to provide the symbol, then it needs a more explanatory commit message than "make lld happy".

stephensmalley:
Ok, passing -Wl,--undefined-version is fine with me.

stephensmalley:
If at first you don't succeed, try try again:
https://lore.kernel.org/selinux/20260611130351.15988-1-stephen.smalley.work@gmail.com/T/#u

stephensmalley:
Per discussion in https://lore.kernel.org/selinux/871pe7n3w3.fsf@redhat.com/T/#t , yet another approach was submitted at: https://lore.kernel.org/selinux/20260615172736.45712-1-stephen.smalley.work@gmail.com/T/#u

PR #516, opened 2026-04-11
adjust audit2why map for python2 and 3

### LINKED ISSUE #461: libselinux 3.7 doesn't compile with ld.lld
```
ld.lld: error: version script assignment of 'AUDIT2WHY_2.9' to symbol 'initaudit2why' failed: symbol not defined
clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: command '/usr/lib/ccache/bin/clang' failed with exit code 1
make[1]: *** [Makefile:151: pywrap] Error 1
make[1]: Leaving directory '/var/tmp/portage/sys-libs/libselinux-3.7-r1/work/libselinux-3.7-abi_x86_64.amd64/src'
```
I encountered [this gentoo bug](https://bugs.gentoo.org/926520) when building a system with LLVM + systemd + linux. I think these issues are supposed to be resolved upstream and not through workarounds like -Wl,--undefined-version - and I have no idea what the correct fix in the code is so I'm posting this issue here.

peppergrayxyz:
align conditional python symbols creation and export.

Different symbols are defined for python2 and python3 in `audit2why.c`:
https://github.com/SELinuxProject/selinux/blob/6be1ec3792c11040fd7a3ecb1135e54418eb0d57/libselinux/src/audit2why.c#L456
https://github.com/SELinuxProject/selinux/blob/6be1ec3792c11040fd7a3ecb1135e54418eb0d57/libselinux/src/audit2why.c#L470-L475

Replicate this structure in `setup.py` and use individual maps for  python2 and python3.

fix: #461

### DIFF
--- a/libselinux/src/audit2why-py2.map
+++ b/libselinux/src/audit2why-py2.map
@@ -1,0 +1,5 @@
+AUDIT2WHY_2.9 {
+  global:
+    initaudit2why;
+  local: *;
+};
--- a/libselinux/src/audit2why-py3.map
+++ b/libselinux/src/audit2why-py3.map
@@ -1,0 +1,5 @@
+AUDIT2WHY_2.9 {
+  global:
+    PyInit_audit2why;
+  local: *;
+};
--- a/libselinux/src/setup.py
+++ b/libselinux/src/setup.py
@@ -1,11 +1,21 @@
-#!/usr/bin/python3
+#!/usr/bin/env python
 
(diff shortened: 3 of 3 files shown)


WavyEbuilder:

bachradsusi:
Python 2 is not supported in this project since SELinux userspace release 20191204 / 3.0
Python 2 EOL was January 2020.

PR #522, opened 2026-06-09
libselinux: avoid out-of-bounds access on zero-length lines

netliomax25-code:
1. find_default_type(), matchmediacon(), customizable_init(), and get_field() each read a line with fgets and then index buf[strlen(buf) - 1] to drop the trailing newline.
2. fgets stores a line that begins with a NUL byte, but strlen() then returns 0, so the index becomes buf[(size_t)-1], reading and (for the first three) writing one byte before the stack buffer.
3. Added a non-empty check before the index, the same guard already used in get_ordered_context_list() (get_context_list.c).

Confirmed under ASan with a default_type line beginning with a NUL byte: before, stack-buffer-underflow on the buf[strlen(buf) - 1] read; after, the line is skipped cleanly.

### DIFF
--- a/libselinux/src/get_default_type.c
+++ b/libselinux/src/get_default_type.c
@@ -38,7 +38,7 @@ static int find_default_type(FILE *fp, const char *role, char **type)
 			errno = EINVAL;
 			return -1;
 		}
-		if (buf[strlen(buf) - 1])
+		if (buf[0] && buf[strlen(buf) - 1])
 			buf[strlen(buf) - 1] = 0;
 
 		ptr = buf;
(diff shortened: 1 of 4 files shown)


stephensmalley:
Patches should be submitted to selinux@vger.kernel.org via git send-email; if you are unable to do so, let us know and we will submit for you.

stephensmalley:
Also, your signed-off-by line must contain your actual name.

stephensmalley:
I have submitted this to the selinux@vger.kernel.org mailing list at:
https://lore.kernel.org/selinux/20260610122735.5486-1-stephen.smalley.work@gmail.com/T/#u

stephensmalley:
Merged via https://github.com/SELinuxProject/selinux/commit/1557ae212747db47cf6c6b664d8e6959f89b4e80

PR #523, opened 2026-06-10
sestatus: null-terminate process and file check entries

netliomax25-code:
Repro: a `[process]` or `[files]` entry on the final line of /etc/sestatus.conf with no trailing newline.
Cause: load_checks() stores each entry with malloc(buf_len) then memcpy(dst, bufp, buf_len), so the buffer has no byte for a terminator and the copy is not null-terminated.
Fix: allocate buf_len + 1 and write the terminator at [buf_len] for both the process and file lists, since the entries are later read as C strings by pidof(), lgetfilecon(), lstat() and printf_tab().

### DIFF
--- a/policycoreutils/sestatus/sestatus.c
+++ b/policycoreutils/sestatus/sestatus.c
@@ -135,22 +135,24 @@ static void load_checks(char *pc[], int *npc, char *fc[], int *nfc)
 				case 0:
 					if (*npc >= MAX_CHECK)
 						break;
-					pc[*npc] = (char *)malloc((buf_len) *
+					pc[*npc] = (char *)malloc((buf_len + 1) *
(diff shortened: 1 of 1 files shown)


stephensmalley:
Patches should be submitted to selinux@vger.kernel.org via git send-email; if you are unable to do so, let us know and we will submit for you. Also, please include your actual name in your signed off by line.

stephensmalley:
Submitted with corrected Signed-off-by line at:
https://lore.kernel.org/selinux/20260610124229.5628-1-stephen.smalley.work@gmail.com/T/#u

stephensmalley:
Merged via https://github.com/SELinuxProject/selinux/commit/5015d50be0434c765aae22fe2bd7011a2629734b

PR #524, opened 2026-06-10
Ci refactor build

pebenito:
This patch set refactors the build step into a custom GH action.  This allows other projects (refpolicy, setools, etc.) to reuse these steps in their CI pipelines.

Example run with this set:
https://github.com/pebenito/selinux/actions/runs/27028226461

For an example use, I've made this change in the SETools CI pipeline. 
Then the SETools build that is part of the refpolicy CI simply calls 
SETools's build action:

https://github.com/SELinuxProject/refpolicy/blob/28ee79d5c811c75fb5ebef9d2c566f58a7c53cc2/.github/workflows/build-setools.yml#L56-L64

Mail list posting:
* https://lore.kernel.org/selinux/20260605171825.362853-1-chpebeni@linux.microsoft.com/T/#t

### DIFF
--- a/.github/actions/build-userspace/action.yml
+++ b/.github/actions/build-userspace/action.yml
@@ -33,6 +33,7 @@ runs:
     - name: Install dependencies
       shell: bash
       run: |
+        echo "::group::Install dependencies"
         sudo apt-get update -q
         sudo apt-get install -qy --no-install-recommends \
             bison \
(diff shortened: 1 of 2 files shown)


stephensmalley:
I was hoping @bachradsusi and/or @jwcart2  would weigh in on these since I am less familiar with the GitHub CI.

pebenito:
Not sure why the test farm ones are failing. I didn't change those.

stephensmalley:
Those only successfully run on pushes to SELinuxProject/selinux itself - they require an API key.

jwcart2:
I pushed a test branch yesterday and everything passed (check format only runs on main, so it didn't run), so it looks good to me. I was planning on ack'ing the patches today.

williamcroberts:
The one thing to consider, and I have had to do this quite a few times in my career, is that CI systems change. I've had to switch systems as previously utiliized, free to open source, build systems have moved to paid services. Some have simply vanished too. So this tightly couples your project to the CI.

What I like to do, is put as much CI logic into something like a bash script. It's important to note that language doesn't matter, but bash is a pervasive target, so it makes a good choice. By doing this, it provides me the following properties:
1. minimal setup means I can move CI systems quickly, Github, Gitlab, Jenkins, etc.
2. Local reproduction is simplified, I can just run the script locally in a matched environment
    a. note that this can be a little bit tricky if they CI provider doesn't provide images. I'll just dump all the installed packages, grab a base image for the OS, and then apply those packages.

Just my two cents, the more you tightly integrate into a CI system, the more your trapped by said CI system.

stephensmalley:
Merged via https://github.com/SELinuxProject/selinux/commit/8f539d94d64713c35b13c1d4032541a9386f5c8a

PR #525, opened 2026-06-13
libsepol: cast to unsigned char in ctype calls

netliomax25-code:
Repro: feed a binary policy module or CIL source whose module name, seusers/file_contexts line, identifier, or IP token holds a byte >= 0x80 to the module-to-CIL or CIL parsing paths, built where char is signed (the x86/ARM default).
Cause: tokenize_str(), get_line(), the *_to_cil()/fix_module_name() helpers, cil_gen_nodecon(), cil_fill_ipaddr(), the AST writer, and cil_verify_name() pass a plain char to isspace/isalnum/isalpha/isdigit, so a high-bit byte sign-extends to a negative int, outside the unsigned char or EOF domain those functions accept (C11 7.4p1).
Fix: cast each argument to unsigned char at the call sites in util.c, module_to_cil.c, cil_build_ast.c, cil_write_ast.c and cil_verify.c, matching the cast already used across libselinux.

### DIFF
--- a/libsepol/cil/src/cil_build_ast.c
+++ b/libsepol/cil/src/cil_build_ast.c
@@ -4507,7 +4507,7 @@ int cil_gen_nodecon(struct cil_db *db, struct cil_tree_node *parse_current,
 	} else {
 		char *addr = parse_current->next->data;
 		if (strchr(addr, ':') ||
-		    (strchr(addr, '.') && isdigit(addr[0]))) {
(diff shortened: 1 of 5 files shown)


stephensmalley:
Amended to use your actual name and submitted at: https://lore.kernel.org/selinux/20260615122840.42032-1-stephen.smalley.work@gmail.com/T/#u

netliomax25-code:
Appreciate you sending it up to the list. The amended version looks right to me; I'll keep an eye on the thread there in case anything comes back from review.

PR #530, opened 2026-06-19
checkpolicy: fix xperm complement at range boundaries

netliomax25-code:
1. avrule_omit_xperms() builds the complement of an omitted ("~") extended permission range set and advances the next allowed range start with r->range.high + 1 kept in a uint16_t.
2. when an omitted range ends at 0xffff that increment wraps to 0, so a spurious [0,0xffff] range is appended and the omitted value ends up granted; an omit range starting at 0 with no successor leaves the trailing range high unset, producing a malformed low > high range.
3. reworked it to walk the sorted, merged ranges with a uint32_t cursor and emit each gap, so the 0 and 0xffff boundaries are handled.

Verified with the roundtrip test: `allowxperm ... ioctl ~0xffff` previously read back as `{ 0x0-0xffff }` (0xffff granted) and now reads back as `{ 0x0-0xfeff }` `{ 0xff00-0xfffe }`. Added regression cases for ~0xffff and ~0x0 to tests/policy_allonce.conf.


checkpolicy/policy_define.c
```diff
@@ -2092,48 +2092,59 @@ static int avrule_read_xperm_ranges(struct av_xperm_range_list **rangehead)
...
+			else
+				newhead = rnew;
+			tail = rnew;
+		}
+		next_low = (uint32_t)r->range.high + 1;
 	}
 
-	while (r) {
-		r2->range.high = r->range.low - 1;
+	if (next_low <= 0xffff) {
```
williamcroberts: Use the constant macro here UINT16_MAX

checkpolicy/policy_define.c
```diff
@@ -2092,48 +2092,59 @@ static int avrule_read_xperm_ranges(struct av_xperm_range_list **rangehead)
...
-		r2 = r2->next;
-
-		r2->range.low = r->range.high + 1;
-		if (!r->next)
-			r2->range.high = 0xffff;
-		r = r->next;
+		rnew->range.low = next_low;
+		rnew->range.high = 0xffff;
+		if (tail)
+			tail->next = rnew;
```
williamcroberts, line 2127: These look like weird formatting issue or something weird with the way GitHub is displaying them, can you check that it's indented properly and there are no whitespace errors reported by git.

williamcroberts (changes requested):

checkpolicy/policy_define.c
netliomax25-code: Done, switched both this comparison and the trailing range.high to UINT16_MAX.

checkpolicy/policy_define.c
netliomax25-code: Checked it - those lines are tab-indented like the rest of the function, and both git diff --check and git show --check report no whitespace errors. Looks like GitHub just rendered the tabs oddly in the diff view.

williamcroberts: Thanks for checking, I appreciate it

stephensmalley:

netliomax25-code:
Sorry about that. My git config had the GitHub handle set as the author name, which is why the Signed-off-by came out wrong. Fixed it locally and amended the commit, so the author and Signed-off-by now both read Kartik Kenchi <netliomax25@gmail.com>. Pushed the corrected version to the branch.

PR #531, opened 2026-06-19
checkpolicy: reject out-of-range extended permission values

netliomax25-code:
avrule_read_xperm_ranges() parses each ioctl/nlmsg command with strtoul() and stores it through a uint16_t cast without checking the range, so a value above 0xffff is silently truncated and a rule is compiled for an unintended command (0x10001 becomes 0x1, 0x123456 becomes 0x3456). The secilc CIL compiler already rejects values outside 0x0000-0xffff. Validate the parsed value here too, before the narrowing cast.

### DIFF
--- a/checkpolicy/policy_define.c
+++ b/checkpolicy/policy_define.c
@@ -2035,6 +2035,19 @@ static void avrule_merge_xperms(struct av_xperm_range_list **rangehead)
 	}
 }
 
+static int avrule_read_xperm_value(const char *id, uint16_t *value)
+{
+	unsigned long val = strtoul(id, NULL, 0);
+
+	if (val > UINT16_MAX) {
+		yyerror2("extended permission value 0x%lx exceeds the maximum of 0x%x",
+			 val, UINT16_MAX);
+		return -1;
+	}
+	*value = (uint16_t)val;
+	return 0;
+}
+
 static int avrule_read_xperm_ranges(struct av_xperm_range_list **rangehead)
 {
 	char *id;
@@ -2053,7 +2066,10 @@ static int avrule_read_xperm_ranges(struct av_xperm_range_list **rangehead)
 			/* high value of range */
 			free(id);
 			id = queue_remove(id_queue);
-			r->range.high = (uint16_t)strtoul(id, NULL, 0);
+			if (avrule_read_xperm_value(id, &r->range.high)) {
+				free(id);
+				return -1;
+			}
 			if (r->range.high < r->range.low) {
 				yyerror2(
 					"extended permission range %#x-%#x must be in ascending order.",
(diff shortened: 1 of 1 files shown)


checkpolicy/policy_define.c
```diff
@@ -2053,16 +2054,33 @@ static int avrule_read_xperm_ranges(struct av_xperm_range_list **rangehead)
...
+					value);
+				free(id);
+				return -1;
+			}
+			r->range.high = (uint16_t)value;
 			if (r->range.high < r->range.low) {
 				yyerror2(
 					"extended permission range %#x-%#x must be in ascending order.",
 					r->range.low, r->range.high);
+				free(id);
```
williamcroberts: Separate out to another patch

checkpolicy/policy_define.c
```diff
@@ -2053,16 +2054,33 @@ static int avrule_read_xperm_ranges(struct av_xperm_range_list **rangehead)
 			/* high value of range */
 			free(id);
 			id = queue_remove(id_queue);
-			r->range.high = (uint16_t)strtoul(id, NULL, 0);
+			value = strtoul(id, NULL, 0);
```
williamcroberts: Use the UINT16_MAX macro here and elsewhere

checkpolicy/policy_define.c
```diff
@@ -2053,16 +2054,33 @@ static int avrule_read_xperm_ranges(struct av_xperm_range_list **rangehead)
...
 		} else {
 			/* read in new low value */
+			value = strtoul(id, NULL, 0);
+			if (value > 0xffff) {
+				yyerror2(
+					"extended permission value 0x%lx exceeds the maximum of 0xffff",
+					value);
+				free(id);
+				return -1;
+			}
```
williamcroberts: I think a small helper for this too would be better than repeating blocks of code.

It could take the Id and a pointer to the uint16_t and return success/fail.

williamcroberts (changes requested):

checkpolicy/policy_define.c
netliomax25-code: Done. Pulled it into a small helper, avrule_read_xperm_value(id, &value), that validates and returns success/fail. Both the low and high sites just call it now, so the block isn't duplicated.

checkpolicy/policy_define.c
netliomax25-code: Switched to UINT16_MAX, and the bound now lives in the helper so there's a single check.

checkpolicy/policy_define.c
netliomax25-code: Dropped it from this PR — the ascending-order path is back to its original behavior here. That free is a pre-existing leak on that branch, so I'll send it as its own patch.

netliomax25-code:
Pushed an update covering all three: the validation moved into a small avrule_read_xperm_value() helper using UINT16_MAX, both call sites use it, and I dropped the unrelated free() on the ascending-order path (will send that one separately).

stephensmalley:

netliomax25-code:
Sorry about that. My local git config still had the GitHub handle instead of my name, so the sign-off came out as netliomax25-code. I've fixed the config and force-pushed an amended commit with `Signed-off-by: Kartik Kenchi <netliomax25@gmail.com>`, so the branch now matches what you posted to the list. Should stay correct on future patches.

PR #532, opened 2026-06-20
libsepol: bound type values in type_set_expand negset loop

netliomax25-code:
Repro: read a base or module policy whose role negset ebitmap has a bit at or above the declared type count (p_types.nprim).
Cause: the negset loop in type_set_expand indexes type_val_to_struct[i] without the nprim bound the sibling types loop in the same function already applies; this expansion runs from policydb_role_cache during policydb_read, before policydb_validate.
Fix: apply the same nprim bound to the negset loop before the dereference.

### DIFF
--- a/libsepol/src/expand.c
+++ b/libsepol/src/expand.c
@@ -2783,6 +2783,13 @@ int type_set_expand(type_set_t *set, ebitmap_t *t, policydb_t *p,
 	/* Now do the same thing for negset */
 	ebitmap_init(&neg_types);
 	ebitmap_for_each_positive_bit(&set->negset, tnode, i) {
+		/*
+		 * invalid policies might have more types set in the ebitmap than
+		 * what's available in the type_val_to_struct mapping
+		 */
+		if (i >= p->p_types.nprim)
+			goto err_neg;
+
 		if (p->type_val_to_struct[i] &&
 		    p->type_val_to_struct[i]->flavor == TYPE_ATTRIB) {
 			if (ebitmap_union(&neg_types,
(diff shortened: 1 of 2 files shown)


williamcroberts:
Did this patch hit the mailing list, I don't see it in the archiver?

We should probably add a test to prevent regressions.

netliomax25-code:
1. Mailing list: no, I haven't sent this one to selinux@vger.kernel.org myself yet. The earlier fixes got picked up straight from the PRs so I left this the same way, but I can send it via git send-email if that's easier on your end.

2. Test: added one in `libsepol/tests/test-expander.c` (`expander_negset_bounds`) and pushed it to the branch. It builds a minimal policydb with nprim=1 and a type_set whose negset has a bit one past the last valid type value, then calls `type_set_expand(..., alwaysexpand=1)`. Without the guard that bit indexes `type_val_to_struct` out of bounds (ASAN reports a heap read overflow in the negset loop); with it the call returns -1, which the test asserts.

williamcroberts:
> 1. Mailing list: no, I haven't sent this one to [selinux@vger.kernel.org](mailto:selinux@vger.kernel.org) myself yet. The earlier fixes got picked up straight from the PRs so I left this the same way, but I can send it via git send-email if that's easier on your end.

I think @stephensmalley was being nice and posted your patches for you to the list. However, it seems like you have a few patches in the queue, it's probably best to follow the CONTRIBUTING.md file.

stephensmalley:
I was out of office yesterday and won't get around to these until Monday. note that if you submit them yourself they still need your real name in the Signed-off-by line.

netliomax25-code:
1. Got it, I'll follow CONTRIBUTING.md and send these to selinux@vger.kernel.org with git send-email instead of leaning on the PRs. I'll get the ones already in the queue onto the list so they're not scattered around.

2. Noted on the Signed-off-by, I'll make sure the name is right on the list submission so it doesn't need rewriting again. No rush, Monday works.

stephensmalley:

PR #533, opened 2026-06-20
mcstrans: fix out-of-bounds read in parse_raw sensitivity parsing

netliomax25-code:
Repro: send mcstransd a context whose range high level is "s-1" (e.g. the range "s0-s-1"); trans_context() splits the range on its first dash and calls parse_raw("s-1").

Cause: parse_raw() advances the cursor with raw += numdigits(sens) + 1 after sscanf(raw, "s%u", &sens). sscanf %u accepts a leading minus, so "s-1" stores UINT_MAX into sens and numdigits() returns 10 while only two characters were consumed. raw moves 11 bytes past the 4-byte allocation and the next *raw == : test reads out of bounds. The string is attacker-controlled: it is the MLS range of any context handed to the mcstransd socket via trans_context()/untrans_context() -> extract_range() -> compute_trans_from_raw(). parse_category() advances with the same numdigits construct at two sites; those are currently shielded only by the >= MAX_CATS checks.

Fix: advance by the number of characters sscanf actually consumed via %n at all three sites and drop numdigits(). Validated with an ASAN build of the parse_raw sensitivity logic: input "s-1" reports a heap-buffer-overflow read 7 bytes past the 4-byte buffer before the change, and runs clean after; "s5", "s5:c1.c3" and "s0" parse unchanged.

### DIFF
--- a/mcstrans/src/mcstrans.c
+++ b/mcstrans/src/mcstrans.c
@@ -173,41 +173,35 @@ err:
 	return -1;
 }
 
-static int numdigits(unsigned int n)
-{
-	int count = 1;
-	while ((n = n / 10))
-		count++;
-	return count;
-}
-
 static int parse_category(ebitmap_t *e, const char *raw, bool allowinverse,
 			  bool fromconfig)
 {
 	int inverse = 0;
 	unsigned int low, high;
 	while (*raw) {
+		int n = 0;
 		if (allowinverse && *raw == '~') {
 			inverse = !inverse;
 			raw++;
 			continue;
 		}
-		if (sscanf(raw, "c%u", &low) != 1)
+		if (sscanf(raw, "c%u%n", &low, &n) != 1)
 			return -1;
 		if (low >= MAX_CATS)
 			return -1;
 		if (!fromconfig && low >= maxbit)
 			return -1;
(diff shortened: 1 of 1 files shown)


williamcroberts (commented):
Reviewed not tested, but it looks good. I had no idea that sscand supports %n.

This needs to go through mailing list like all patches and as @stephensmalley mentioned, your signoff and author details need to be your name.

Negative diffstat for the win, my favorite kind of patches.

netliomax25-code:
Sounds good.

1. The commit already has the full description in the body and a Signed-off-by, so it should be ready as-is. The author and sign-off both use the handle I contribute under; let me know if you need the From/sign-off in a different form for the list and I'll adjust.

2. On the not-tested point: I validated the sensitivity path under ASAN before sending. "s-1" reads 7 bytes past the 4-byte buffer before the change and runs clean after, while "s5", "s5:c1.c3" and "s0" parse unchanged. Can attach the standalone repro if it's useful.

3. I'm not currently set up for git send-email. You'd offered before to post on a contributor's behalf, so if that still stands I'd appreciate it; otherwise I'll get send-email working and submit it myself.

stephensmalley:

PR #534, opened 2026-06-21
libsepol/cil: fix double free of borrowed type datum on error path

netliomax25-code:
1. cil_typepermissive_to_policydb() and cil_typeneveraudit_to_policydb() set sepol_type from __cil_get_sepol_type_datum(), which is a hashtab_search() into pdb->p_types, so the datum is owned by the policydb.
2. their error path (taken when ebitmap_set_bit on permissive_map/neveraudit_map fails) calls type_datum_destroy(sepol_type) and free(sepol_type), and the same type is then freed again by type_destroy() during policydb_destroy(), a double free.
Removed the destroy and free of the borrowed datum from both error paths. The sibling alias converters (cil_typealias/cil_catalias/cil_sensalias_to_policydb) already follow this rule, freeing only their own allocation and leaving the lookup result alone.

### DIFF
--- a/libsepol/cil/src/cil_binary.c
+++ b/libsepol/cil/src/cil_binary.c
@@ -611,18 +611,13 @@ int cil_typepermissive_to_policydb(policydb_t *pdb,
 	rc = __cil_get_sepol_type_datum(pdb, DATUM(cil_typeperm->type),
 					&sepol_type);
 	if (rc != SEPOL_OK)
-		goto exit;
+		return rc;
 
 	if (ebitmap_set_bit(&pdb->permissive_map, sepol_type->s.value, 1)) {
-		goto exit;
+		return rc;
 	}
 
 	return SEPOL_OK;
-
-exit:
-	type_datum_destroy(sepol_type);
-	free(sepol_type);
-	return rc;
 }
 
 int cil_typeneveraudit_to_policydb(policydb_t *pdb,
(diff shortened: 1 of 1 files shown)


libsepol/cil/src/cil_binary.c
```diff
@@ -620,8 +620,6 @@ int cil_typepermissive_to_policydb(policydb_t *pdb,
 	return SEPOL_OK;
 
 exit:
-	type_datum_destroy(sepol_type);
-	free(sepol_type);
 	return rc;
```
williamcroberts: If the label isn't doing anything, then it can probably be short circuited by replacing the goto with a return rc.

libsepol/cil/src/cil_binary.c
```diff
@@ -643,8 +641,6 @@ int cil_typeneveraudit_to_policydb(policydb_t *pdb,
 	return SEPOL_OK;
 
 exit:
-	type_datum_destroy(sepol_type);
-	free(sepol_type);
```
williamcroberts, line 647: Same here

williamcroberts (changes requested):

libsepol/cil/src/cil_binary.c
netliomax25-code: Makes sense. With the destroy/free gone the label was only returning rc, so I dropped the goto/label and return directly in both spots. The ebitmap path still returns rc to keep the original behavior. Pushed.

libsepol/cil/src/cil_binary.c
netliomax25-code: Done here too.

stephensmalley:

netliomax25-code:
Sorry for the repeated hassle on the sign-off. I'd seen you already corrected it and posted to the list, so I didn't re-push over the PR branch, figuring the list copy was the canonical one. I've fixed my git config so the author line and the Signed-off-by both come from netliomax25-code <netliomax25@gmail.com> and stay consistent on future patches. That's the name I contribute under, so that's what ends up in the sign-off. Thanks for taking it over.

stephensmalley:
We want your real name in the Signed-off-by line since the DCO is a legal statement wrt copyright/license of the code and isn't meaningful if just a "handle" or "nickname" is used. At least that's the perspective of the Linux kernel folks, not sure what folks on GitHub accept these days.

netliomax25-code:
ALNIDA KHAN

stephensmalley:
Sorry, I'm confused - on one of the earlier commits you used Kartik Kenchi so that is what I have been using for the patch postings.

netliomax25-code:
yeah you can use that also

stephensmalley:
Ok, I will keep using Kartik Kenchik for consistency with your prior patches unless/until you ask me to do otherwise.

netliomax25-code:
ok thanks

